PR #18, opened 2019-01-23
Update/unified video player

devgabrielcoman:
## JIRA
No formal JIRA link, but it's related to the iOS ticket; it's basically making sure both Android & iOS are using as much common architecture as possible; 
[KWS-5180](https://jira.superawesome.tv/browse/KWS-5180)

## Description
a) Rewritten or amended the `SAVideoAd` class; a class full of static methods; it's simpler now; 
b) Split out eventing for video into `SAVideoEvents` and `SAVideoClick`
c) The activity that plays the video now only gets injected with some config objects (an `SAAd` instance, an `SAEvents` instance, etc) and builds internally what it needs (`video player`, `media control` ,etc)

a + b + c ==> less shared state between objects, less head-aches for me.


superawesome-base/build.gradle
```diff
@@ -91,12 +96,42 @@ repositories {
...
+
+    compile "tv.superawesome.lib.sautils:sautils:$saUtils"
+    compile "tv.superawesome.lib.saevents:saevents:$saEvents"
+    compile "tv.superawesome.lib.sawebplayer:sawebplayer:$saWebPlayer"
+    compile "tv.superawesome.lib.saadloader:saadloader:$saAdLoader"
+    compile "tv.superawesome.lib.saparentalgate:saparentalgate:$saParentalGate"
+    compile "tv.superawesome.lib.sabumperpage:sabumperpage:$saBumperPage"
+    compile "tv.superawesome.sagdprisminorsdk:sagdprisminorsdk:$saIsMinor"
+//    compile "tv.superawesome.lib.savideoplayer:savideoplayer:$saVideoPlayer"
+    compile fileTree(dir: '../../sa-mobile-lib-android-videoplayer/savideoplayer/build/outputs/aar', include: ['*.aar'])
```
GuilhermeMota93: This doesn't seem right...

devgabrielcoman: As soon as video is updated, I'll revert to gradle dependencies

PR #22, opened 2020-05-20
Implementation of the new networking layer

gunhansancar:


build.gradle
```diff
@@ -21,3 +28,10 @@ allprojects {
         maven { url "https://maven.google.com" }
     }
 }
+
+ext.getSdkVersion = { ->
```
devgabrielcoman, line 32: Woot 👍 @GuilhermeMota93 nice touch, check this out!

settings.gradle
```diff
@@ -1,2 +1,5 @@
+include ':superawesome-common'
+include ':superawesome-retrofit'
+include ':superawesome-core'
 include ':app', ':saunity', ':samopub', ':saair', ':superawesome', ':superawesome-base'
```
devgabrielcoman, line 4: might want to put all of the other "include" module statements on their own line, just for consistency.

superawesome-common/build.gradle
```diff
@@ -0,0 +1,55 @@
...
+    id 'kotlin-android'
+    id 'kotlin-android-extensions'
+}
+
+android {
+    compileSdkVersion 29
+    buildToolsVersion "29.0.3"
+
+    defaultConfig {
+        minSdkVersion 16
```
devgabrielcoman, line 13: I kindof feel like all the values here:

```
minSdkVersion 16
targetSdkVersion 29
.....
implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.6"
...
```

all of these values: 16, 29, 1.3.6 - need to be defined in a central gradle file and accessed from inside each of the project files. 

otherwise it's going to become progressively harder to keep things in sync, even when we control all of these things.

superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/Device.kt
```diff
@@ -0,0 +1,35 @@
...
+            val widthDp = width.toDouble() / density.toDouble()
+            val heightDp = height.toDouble() / density.toDouble()
+            return sqrt(widthDp.pow(2.0) + heightDp.pow(2.0))
+        }
+
+    private val deviceCategory: DeviceCategory
+        get() = if (systemSize <= PHONE_SCREEN_MAX) DeviceCategory.phone else DeviceCategory.tablet
+
+    companion object {
+        const val PHONE_SCREEN_MAX = 6.2
```
devgabrielcoman, line 33: maybe make this `private`

superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/Encoder.kt
```diff
@@ -0,0 +1,21 @@
...
+    fun encodeUri(string: String?): String
+}
+
+class Encoder : EncoderType {
+    private object Keys {
+        const val utf8 = "UTF-8"
+    }
+
+    override fun encodeUri(string: String?): String =
+            if (string.isNullOrEmpty()) "" else try {
```
devgabrielcoman, line 16: might want to add a proper if/else w/ curly brackets here .... it's a little bit confusing. Sometimes Kotlin's syntax can be too sweet.

superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/GoogleAdvertisingProxy.kt
```diff
@@ -0,0 +1,37 @@
+package tv.superawesome.sdk.publisher.common.components
+
+import android.content.Context
+import android.util.Log
+
+interface GoogleAdvertisingProxyType {
+    suspend fun findAdvertisingId(): String?
+}
+
+class GoogleAdvertisingProxy(private val context: Context) : GoogleAdvertisingProxyType {
```
devgabrielcoman, line 10: Nice name for this 👍

superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/GoogleAdvertisingProxy.kt
```diff
@@ -0,0 +1,37 @@
...
+    private object Keys {
+        const val GOOGLE_ADVERTISING_CLASS = "com.google.android.gms.ads.identifier.AdvertisingIdClient"
+        const val GOOGLE_ADVERTISING_ID_CLASS = "com.google.android.gms.ads.identifier.AdvertisingIdClient\$Info"
+        const val GOOGLE_ADVERTISING_INFO_METHOD = "getAdvertisingIdInfo"
+        const val GOOGLE_ADVERTISING_TRACKING_METHOD = "isLimitAdTrackingEnabled"
+        const val GOOGLE_ADVERTISING_ID_METHOD = "getId"
+    }
+
+    override suspend fun findAdvertisingId(): String? {
```
devgabrielcoman, line 19: The method is suspending, but are all of these google method **suspending** as well? cause otherwise I'd bet they're thread **blocking**.

Probably the thing to do here is: 

```
override suspend fun findAdvertisingId(): String? {
   return withContext(Dispatchers.IO) { 
       try { ... } catch (exception: Exception) { ... }
   }
}
```
 
notice the `withContext` call which takes us from non-suspending & blocking to suspending.

superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/UserAgentProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+import android.webkit.WebSettings
+import android.webkit.WebView
+
+interface UserAgentProviderType {
+    val name: String
+}
+
+class UserAgentProvider(context: Context) : UserAgentProviderType {
+    override val name: String by lazy {
+        if (Build.VERSION.SDK_INT >= 17) {
```
devgabrielcoman, line 14: Should 17 be hardcoded here? Can we use `Build.VERSION_CODES.XXX`?

PR #25, opened 2020-06-18
Fix/mopub part 1

gunhansancar:
Upgrading mopub to the latest version


app/src/main/java/tv/superawesome/demoapp/MoPubActivity.kt
```diff
@@ -0,0 +1,151 @@
...
+    private var banner: MoPubView? = null
+    private var interstitial: MoPubInterstitial? = null
+
+    companion object {
+        private const val BannerId = "57e4e6e7cf784086b733bb4a427b8020"
+        private const val InterstitialId = "947fe751ce2a4a56802e882e1d902972"
+        private const val VideoId = "89c0bc90a4a545d99f59429a67f838b9"
+
+        // MoPub Test Ids
+//        private const val BannerId = "b195f8dd8ded45fe847ad89ed1d016da"
```
devgabrielcoman, line 24: I suppose you can delete these?

samopub/src/main/java/tv/superawesome/plugins/publisher/mopub/AwesomeAdsMoPubAdDataExtractor.kt
```diff
@@ -0,0 +1,115 @@
+package tv.superawesome.plugins.publisher.mopub
+
+import com.mopub.mobileads.AdData
+import tv.superawesome.lib.sasession.defines.SAConfiguration
+import tv.superawesome.sdk.publisher.SADefaults
+import tv.superawesome.sdk.publisher.SAOrientation
+
+class AwesomeAdsMoPubAdDataExtractor(private val adData: AdData) {
```
devgabrielcoman, line 8: Am I reading this wrong - or do we have this class defined twice?

gunhansancar: Yes, you're right. When I changed the package names it seems the new classes are duplicated.
Removed them.

PR #31, opened 2020-08-11
Update gradle files

gunhansancar:
-Removed deprecated `com.github.dcendents:android-maven-gradle-plugin` plugin
-Updated publishing to the central location into `bintray.gradle`
-Updated all `build.gradle` files in each project

### DIFF
--- a/bintray.gradle
+++ b/bintray.gradle
@@ -73,8 +73,9 @@ bintray {
     publications = ['MyPublication']
 
     pkg {
-        repo = "maven"
+        repo = "SuperAwesomeSDK"
         name = this.publishName
+        userOrg = 'superawesome'
         websiteUrl = siteUrl
         vcsUrl = gitUrl
         licenses = ["Apache-2.0"]
--- a/bintray.gradle
+++ b/bintray.gradle
@@ -1,0 +1,83 @@
+apply plugin: 'com.jfrog.bintray'
+apply plugin: 'maven-publish'
+
+Properties properties = new Properties()
+properties.load(project.rootProject.file('local.properties').newDataInputStream())
+
+version = properties.getProperty("project.version")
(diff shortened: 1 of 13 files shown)


bintray.gradle
```diff
@@ -0,0 +1,83 @@
...
+        MyPublication(MavenPublication) {
+            pom.withXml {
+                asNode().with {
+                    getAt('packaging')[0].replaceNode { packaging('aar') }
+                    appendNode('name', this.publishDescription)
+                    appendNode('url', siteUrl)
+
+                    appendNode('licenses').with {
+                        appendNode('license').with {
+                            appendNode('name', 'The Apache Software License, Version 2.0')
```
devgabrielcoman, line 33: is this the actual license?

bintray.gradle
```diff
@@ -0,0 +1,83 @@
...
+                            appendNode('url', 'http://www.apache.org/licenses/LICENSE-2.0.txt')
+                        }
+                    }
+
+                    appendNode('developers').with {
```
devgabrielcoman, line 38: if we're doing this, does my name / account have to be here? thinking maybe we transition to an official SA one?

bintray.gradle
gunhansancar: It is what we had so far. Didn't change it. If needed we can update it to something else.

bintray.gradle
gunhansancar: AFAIK developers tag is needed if we publish to the sonatype maven central. But I couldn't find our repo over there which means we don't sync.

It is better to sync to Jcenter (which is superset of maven central) which doesn't have any requirements like developers etc. 

And then we can even remove `maven { url "http://dl.bintray.com/gabrielcoman/maven" }` personal maven repository links.

PR #38, opened 2021-03-09
Add events for moat

nicolassleimanSA:
Track moat events to know how many failure/attempt we have

### DIFF
--- a/samoat/src/main/java/tv/superawesome/plugins/publisher/moat/SAMoatEvents.java
+++ b/samoat/src/main/java/tv/superawesome/plugins/publisher/moat/SAMoatEvents.java
@@ -73,7 +73,11 @@ public class SAMoatEvents implements TrackerListener, VideoTrackerListener {
         Log.d(TAG, "Trying to start tracking web with query " + moatQuery);
 
         // and return the special moat javascript tag to be loaded in a web view
-        return "<script src=\""+MOAT_SERVER+"/"+MOAT_DISPLAY_PARTNER_CODE+"/"+MOAT_URL+"?"+moatQuery+"\" type=\"text/javascript\"></script>";
+        return "<script defer" +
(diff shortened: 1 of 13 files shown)


samoat/src/main/java/tv/superawesome/plugins/publisher/moat/SAMoatEvents.java
```diff
@@ -73,7 +73,11 @@ public String startMoatTrackingForDisplay(WebView webView, HashMap<String, Strin
         Log.d(TAG, "Trying to start tracking web with query " + moatQuery);
 
         // and return the special moat javascript tag to be loaded in a web view
-        return "<script src=\""+MOAT_SERVER+"/"+MOAT_DISPLAY_PARTNER_CODE+"/"+MOAT_URL+"?"+moatQuery+"\" type=\"text/javascript\"></script>";
+        return "<script defer" +
```
devgabrielcoman, line 76: would we need `type='text/javascript'` on the tag?

superawesome-base/src/main/java/tv/superawesome/lib/saevents/SAEvents.java
```diff
@@ -182,6 +181,9 @@ public static void initMoat (Application application, boolean loggingEnabled) {
     }
 
     public String startMoatTrackingForDisplay(WebView view) {
+        if(moatModule == null) {
+            triggerMoatAttemptNoClassEvent();
```
devgabrielcoman, line 185: could probably put

```
if (moatModule == null) {
 triggerMoatAttemptNoClassEvent();
 return "";
}

return moatModule.startMoatTrackingForDisplay(view);
```

if you're okay w/ early returns and don't want to make the same check on line 187 again.

devgabrielcoman (approved):
Looks okay to me.

PR #39, opened 2021-03-25
feat(): Android SDK to load WebSDK for display ads.

ari-ada:
Android SDK to load WebSDK for display ads.

### DIFF
--- a/app/src/main/java/tv/superawesome/demoapp/MainActivity.java
+++ b/app/src/main/java/tv/superawesome/demoapp/MainActivity.java
@@ -29,6 +29,8 @@ import tv.superawesome.sdk.publisher.SAEvent;
 import tv.superawesome.sdk.publisher.SAInterface;
 import tv.superawesome.sdk.publisher.SAInterstitialAd;
 import tv.superawesome.sdk.publisher.SAVideoAd;
+import tv.superawesome.sdk.publisher.managed.SAManagedBannerAd;
+import tv.superawesome.sdk.publisher.managed.SAManagedInterstitialAd;
 
 public class MainActivity extends Activity {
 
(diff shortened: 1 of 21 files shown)


superawesome-base/src/main/java/tv/superawesome/sdk/publisher/managed/SAManagedBannerAd.kt
```diff
@@ -0,0 +1,223 @@
...
+
+    fun JSONObject.toQueryParams(): String {
+        var queryParams = ""
+
+        var looper = this.keys()
+        while(looper.hasNext()) {
+            val key = looper.next()
+            if (key.equals("test") && this.get(key).toString().equals("false")) {
+                continue
+            }
```
ari-ada, line 89: We found a bug in the ads-server. If you pass `false` then ads-server won't return. Will fix the ads-server but those lines won't affect anything else.

ari-ada:
the Activity does not need to be merged. I have prepared it for the green team for testing but it is not needed afterwards.

app/src/main/java/tv/superawesome/demoapp/MainActivity.java
```diff
@@ -39,67 +41,25 @@
...
-        SAVideoAd.enableBackButton();
-        SAVideoAd.setListener((SAInterface) (placementId, event) -> {
-            Log.d("SADefaults", "VIDEO AD: " + placementId + " -> Event : " + event);
-
-            if (event == SAEvent.adLoaded) {
-//                SAVideoAd.play(placementId, MainActivity.this);
-            }
+        final SAManagedBannerAd managedBannerAd = findViewById(R.id.preview);
+        managedBannerAd.setListener((SAInterface) (placementId, event) -> {
+            managedBannerAd.load(placementId);
```
devgabrielcoman, line 48: is this load or play?

build.gradle
```diff
@@ -1,4 +1,5 @@
 buildscript {
+    ext.kotlin_version = '1.3.71'
```
devgabrielcoman, line 2: you're  possibly running two kotlin versions here? stick with the one that says `1.3.72`, it's very stable. Unless you have a reason to switch to `1.3.71`

devgabrielcoman (approved):
Looks okay @ari-ada plus/minus a few comments

ari-ada:
ok. We went through all problems with Gabriel and Nicolas. Will deal with merge conflicts and merge

PR #41, opened 2021-03-30
Migrate refactor to latest gradle

markGilchrist:


superawesome-ui/src/main/java/tv/superawesome/sdk/publisher/ui/video/AdVideoPlayerControllerView.kt
```diff
@@ -105,7 +105,6 @@ constructor(context: Context?, attrs: AttributeSet? = null, defStyleAttr: Int =
         // call to super
 
         // create the background image
```
dnharjani-sa, line 107: Feel free to remove comments like this that address the what and not the why. If it's unclear what a function / var is doing without a comment like this then we should change the name rather than add comments

superawesome-unity/src/main/java/tv/superawesome/plugins/publisher/unity/SAUnityInterstitialAd.kt
```diff
@@ -31,15 +31,15 @@ object SAUnityInterstitialAd {
...
+                    SAEvent.AdLoaded -> SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.AdLoaded.toString())
+                    SAEvent.AdEmpty -> SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.AdEmpty.toString())
+                    SAEvent.AdFailedToLoad -> SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.AdFailedToLoad.toString())
+                    SAEvent.AdAlreadyLoaded -> SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.AdAlreadyLoaded.toString())
+                    SAEvent.AdShown -> SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.AdShown.toString())
+                    SAEvent.AdFailedToShow -> SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.AdFailedToShow.toString())
+                    SAEvent.AdClicked -> SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.AdClicked.toString())
+                    SAEvent.AdEnded -> SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.AdEnded.toString())
+                    SAEvent.AdClosed -> SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.AdClosed.toString())
```
dnharjani-sa, line 42: Just note that the unity bridge might be expecting a different casing of the name and will need to be updated as a result.

dnharjani-sa (approved):
Looks fine, just need to test and make sure that there are no dependencies on the casing of the event enums - probably worth checking in the documentation to be sure

superawesome-ui/src/main/java/tv/superawesome/sdk/publisher/ui/video/AdVideoPlayerControllerView.kt
dnharjani-sa: Perfect

superawesome-unity/src/main/java/tv/superawesome/plugins/publisher/unity/SAUnityInterstitialAd.kt
markGilchrist: I have checked all the function in the unity - bridge and from what I can see the were i no place where a string (from unity ) then converted to SAEvent member.

is there another way I can check this?

dnharjani-sa: I think that's probably enough - the main thing is that the callback functions themselves are unchanged because that's what publishers would rely on

devgabrielcoman: The issue wouldn't be the unity bridge but the C# code in this repo: https://github.com/SuperAwesomeLTD/sa-unity-sdk

devgabrielcoman: If you think there's no issue - then there's no issue!

devgabrielcoman (commented):
Just echoing @dnharjani-sa 's comment about that enum case.

superawesome-unity/src/main/java/tv/superawesome/plugins/publisher/unity/SAUnityInterstitialAd.kt
dnharjani-sa: Mark's working on tests for this I assume so we should catch any problems

PR #93, opened 2021-06-02
Improvement/remove warnings

markGilchrist:


superawesome-base/src/test/java/tv/superawesome/lib/saevents/events/trigger/TestSAClickEventTrigger.java
```diff
@@ -36,11 +33,8 @@ public void test_SAClickEvent_triggerEvent_WithDisplayAdFailure () {
         SAClickEvent event = new SAClickEvent(ad, super.session, super.executor, 1000, true);
 
         // when
-        event.triggerEvent(success -> {
-
-            // then
-            Assert.assertFalse(success);
-        });
+        // then
```
dnharjani-sa, line 36: You can remove these comments everywhere, they're pretty useless

dnharjani-sa (approved):

superawesome-base/src/test/java/tv/superawesome/lib/saevents/events/trigger/TestSAClickEventTrigger.java
dnharjani-sa: Given / When / Then is useful if you're tying them back to plain english feature descriptions. The code is declarative enough in these tests so that following them doesn't require these helper comments.

If this was along the lines of 

// GIven: I have a click event that is missing X
// When: I trigger that click event
// Then: The click event result should be false

That would be fine. As it is though we're better off having this at the top in the actual test description e.g.

describe("Click event with improperly configured value for X")

markGilchrist: I will address this in a future pr for tests that are simple to look at

PR #99, opened 2021-07-15
update Url for line Item

markGilchrist:


superawesome-base/src/main/java/tv/superawesome/lib/saadloader/SALoader.java
```diff
@@ -224,7 +224,7 @@ public void loadAdForLineItem(
           @NonNull final SALoaderInterface listener) {
 
     // get connection things to AwesomeAds
-    String endpoint = session.getBaseUrl() + "/ads/" + lineItemId + "/" + creativeId;
+    String endpoint = session.getBaseUrl() + "/ad/" + lineItemId + "/" + creativeId;
```
dnharjani-sa, line 227: Why do we have 2 different methods in here that do the same thing?

loadAd that takes the creativeid and lineitemId AND loadAdForLineItem that calls the underlying function.

Reason I ask is because SAVideo uses loadAd directly, interstitial and banner use loadAdForLineItem.

markGilchrist: I have amended this as it was clearly an error

dnharjani-sa (approved):

PR #129, opened 2021-11-25
Ensure configuration read and written by same logic

markGilchrist:


dnharjani-sa:
This has changes from a ton of different branches. Can we rebase and only put in the changes that are relevant to the issue.

dnharjani-sa:
> This has changes from a ton of different branches. Can we rebase and only put in the changes that are relevant to the issue.

Also there are failing tests

PR #134, opened 2022-02-01
Aag 1590 test app placements

natashashafqat:
5 working placements have been created and added.

Screenshot 2022-02-01 at 19 02 18

### DIFF
--- a/app/src/main/java/tv/superawesome/demoapp/MainActivity.java
+++ b/app/src/main/java/tv/superawesome/demoapp/MainActivity.java
@@ -47,15 +47,13 @@ public class MainActivity extends Activity {
     final List<AdapterItem> data =
         Arrays.asList(
             new HeaderItem("Banners"),
-            new PlacementItem("320x50 - Mobile Leaderboard", 78149, Type.BANNER),
+            new PlacementItem("Mobile Leaderboard", 78149, Type.BANNER),
             new HeaderItem("Interstitials"),
-            new PlacementItem("Rich Media - Mobile Interstitial Portrait", 78150, Type.INTERSTITIAL),
-            new PlacementItem("Static - Mobile Interstitial Portrait", 61298, Type.INTERSTITIAL),
+            new PlacementItem("Mobile Interstitial Portrait", 78150, Type.INTERSTITIAL),
+            new PlacementItem("Interstitial via KSF", 82063, Type.INTERSTITIAL),
             new HeaderItem("Videos"),
-            new PlacementItem("Video Direct", 44262, Type.VIDEO),
-            new HeaderItem("KSF"),
-            new PlacementItem("Insterstitial - Static", 61321, Type.INTERSTITIAL),
-            new PlacementItem("Insterstitial - Rich Media", 61320, Type.INTERSTITIAL));
+            new PlacementItem("Direct Video", 78148, Type.VIDEO),
+            new PlacementItem("VPAID via KSF", 82064, Type.VIDEO));
     ListAdapter<AdapterItem> adapter = new ListAdapter<>(this);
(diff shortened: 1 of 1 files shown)


app/src/main/java/tv/superawesome/demoapp/MainActivity.java
```diff
@@ -47,16 +47,13 @@ protected void onCreate(final Bundle savedInstanceState) {
...
-            new PlacementItem("Insterstitial - Static", 61298, Type.INTERSTITIAL),
-            new HeaderItem("KSF"),
-            new PlacementItem("Insterstitial - Static", 61321, Type.INTERSTITIAL),
-            new PlacementItem("Insterstitial - Rich Media", 44262, Type.INTERSTITIAL),
-            new PlacementItem("Video", 61320, Type.VIDEO));
+            new PlacementItem("Mobile Interstitial Portrait", 82089, Type.INTERSTITIAL),
+            new PlacementItem("Interstitial via KSF", 82063, Type.INTERSTITIAL),
+            new HeaderItem("Videos"),
+            new PlacementItem("Direct Video", 82090, Type.VIDEO),
+            new PlacementItem("VPAID via KSF", 82064, Type.VIDEO));
```
dnharjani-sa, line 56: I wonder if we want to differentiate between KSF specific types in the future

natashashafqat: Like a separate KSF section? So I did originally do this, but I wanted it to mirror what we'd see in iOS, and a new section wouldn't right now be possible there - iOS does some funky logic choosing which loader it uses based on the sections, so would have to just modify that logic for a new KSF section. But I definitely agree it'd help with usability

gunhansancar (approved):

app/src/main/java/tv/superawesome/demoapp/MainActivity.java
dnharjani-sa: I mean more just in the Type ENUM as there is a difference between an interstitial or video loaded by KSF vs one loaded normally

natashashafqat: Oh gotcha! Yeah, it could be a shout to be able to differentiate. Right now the types are just being used to select the loaders (and we're currently loading normal and KSF placements through the same type loader). Would you suggest KSF specific types mainly for readability or something else?

dnharjani-sa: KSF uses the SA*Managed classes which would probably benefit from different loaders as well. That way we don't overload the functionality of the standard classes.

natashashafqat: Ah right, good to know and makes sense - I can create a refactor ticket if we think this is something we wanna do?

dnharjani-sa: Its on the list of things to address already

gunhansancar: Currently, the example app above does not use the new `SAManaged*` classes. Thus, loading and add for KSF is being done using the existing medium.

For the better user experience, we should internally decide if the placement requires to use `SAManaged*` classes. So, we only expose `SABanner`, `SAInterstitial`, `SAVideo` to the user.

dnharjani-sa: From the public API that's fine as no publisher should have to use the Managed classes. However we want to be able to flag a placement to force the use of he managed classes in certain situations. KSF and VPAID is one of them but there are others. As such we also want to be able to test these atomically. How we accomplish this doesn't really matter to me.

PR #137, opened 2022-02-03
Aag 1591 callback logs

natashashafqat:
Adding debug logs for:
- adLoaded
- adShown
- adClicked
- adEnded
- adClosed

Debug logs can now be seen in Logcat, and can easily be found by keeping a filter on for `Event callback: `.

Video below showing debug logs being registered and easily seen for the above callbacks.

https://user-images.githubusercontent.com/48458132/152332378-b6dee0f5-41da-4c4f-b136-2be1805b9419.mov

### DIFF
--- a/superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SABannerAd.java
+++ b/superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SABannerAd.java
@@ -164,7 +164,7 @@ public class SABannerAd extends FrameLayout {
                     if (listener != null) {
                         SAEvent eventToSend = response.isValid() ? SAEvent.adLoaded : SAEvent.adEmpty;
                         listener.onEvent(placementId, eventToSend);
-                        Log.d("SABannerAd.load", "Event callback: " + eventToSend.toString());
+                        Log.d("SABannerAd.load.PL_ID", "Event callback: " + eventToSend.toString());
                     } else {
                         Log.w("AwesomeAds", "Banner Ad listener not implemented. Event would have been either adLoaded or adEmpty");
                     }
@@ -232,7 +232,7 @@ public class SABannerAd extends FrameLayout {
                     if (listener != null) {
                         SAEvent eventToSend = response.isValid() ? SAEvent.adLoaded : SAEvent.adEmpty;
                         listener.onEvent(placementId, eventToSend);
-                        Log.d("SABannerAd.load", "Event callback: " + eventToSend.toString());
+                        Log.d("SABannerAd.load.ALL_IDs", "Event callback: " + eventToSend.toString());
                     } else {
                         Log.w("AwesomeAds", "Banner Ad listener not implemented. Event would have been either adLoaded or adEmpty");
                     }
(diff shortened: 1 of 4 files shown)


superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SABannerAd.java
```diff
@@ -162,7 +162,9 @@ public void load (final int placementId) {
                     canPlay = response.isValid();
                     setAd(response.isValid() ? response.ads.get(0) : null);
                     if (listener != null) {
-                        listener.onEvent(placementId, response.isValid() ? SAEvent.adLoaded : SAEvent.adEmpty);
+                        SAEvent eventToSend = response.isValid() ? SAEvent.adLoaded : SAEvent.adEmpty;
+                        listener.onEvent(placementId, eventToSend);
+                        Log.d("SABannerAd.load.PL_ID", "Event callback: " + eventToSend.toString());
```
gunhansancar, line 167: Log tags (the first parameter) should be same for everything, or at least the file based, and it should be short. Please let's use the same tag, it is better for filtering as well.

superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SABannerAd.java
```diff
@@ -228,7 +230,9 @@ public void load (final int placementId, final int lineItemId, final int creativ
                     canPlay = response.isValid();
                     setAd(response.isValid() ? response.ads.get(0) : null);
                     if (listener != null) {
-                        listener.onEvent(placementId, response.isValid() ? SAEvent.adLoaded : SAEvent.adEmpty);
+                        SAEvent eventToSend = response.isValid() ? SAEvent.adLoaded : SAEvent.adEmpty;
+                        listener.onEvent(placementId, eventToSend);
+                        Log.d("SABannerAd.load.ALL_IDs", "Event callback: " + eventToSend.toString());
```
gunhansancar, line 235: Same here, I wish we started using the refactored code which contains better logger already

superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SAInterstitialAd.java
```diff
@@ -187,6 +188,7 @@ private void close () {
      * @param placementId   the Ad placement id to load data for
      * @param context       the current context
      */
+    @SuppressLint("LongLogTag")
```
gunhansancar, line 191: Let's not suppress warnings and solve the issue by using the smaller same tags

gunhansancar (changes requested):
Please consider my comments for all the changes as I didn't mark them all

superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SABannerAd.java
natashashafqat: Makes sense 👍  I wasn't sure on how to label the tags. I can just do file name? So in this case, just `SABannerAd`?

gunhansancar (approved):

PR #145, opened 2022-04-06
AAG-1791 Fix ad sounds continue to play even after minimising the app

gnhn:
AAG-1791 Fix ad sounds continue to play even after minimising the app

### DIFF
--- a/superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SAVideoActivity.java
+++ b/superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SAVideoActivity.java
@@ -4,7 +4,6 @@
  */
 package tv.superawesome.sdk.publisher;
 
-import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.content.Intent;
 import android.content.pm.ActivityInfo;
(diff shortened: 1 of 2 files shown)


superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SAVideoActivity.java
```diff
@@ -101,7 +106,7 @@ protected void onCreate(Bundle savedInstanceState) {
         chrome.setClickListener(view -> {
             videoClick.handleAdClick(view);
             listenerRef.onEvent(ad.placementId, SAEvent.adClicked);
-            Log.d("SAVideoActivity", "Event callback: " + SAEvent.adClicked.toString());
+            Log.d("SAVideoActivity", "Event callback: " + SAEvent.adClicked);
```
tashmulla, line 109: So we don't need to use the `toString()` method here?

tashmulla (approved):
Looks good!

superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SAVideoActivity.java
gnhn: Yes, for those `enums` `toString` methods already overridden so it is redundant to call `toString`

PR #146, opened 2022-04-08
add: JS bridge to listen events from the WebView

gnhn:
add: JS bridge to listen events from the WebView


app/src/main/java/tv/superawesome/demoapp/MainActivity.java
```diff
@@ -29,209 +29,208 @@
...
+                });
+
+        ListView myList = findViewById(R.id.MyList);
+        final List<AdapterItem> data =
+                Arrays.asList(
+                        new HeaderItem("Banners"),
+                        new PlacementItem("Mobile Leaderboard", 82088, Type.BANNER),
+                        new HeaderItem("Interstitials"),
+                        new PlacementItem("Mobile Interstitial Portrait", 82089, Type.INTERSTITIAL),
+                        new PlacementItem("Interstitial via KSF", 43306, Type.INTERSTITIAL),
```
tashmulla, line 55: Could we swap these KSF placements back to the Production ones please?

app/src/main/java/tv/superawesome/demoapp/MainActivity.java
```diff
@@ -29,209 +29,208 @@
...
+                                SAInterstitialAd.load(placement.pid, MainActivity.this);
+                                break;
+
+                            case VIDEO:
+                                SAVideoAd.load(placement.pid, this);
+                        }
+                    }
+                });
+
+        SAInterstitialAd.setConfigurationStaging();
```
tashmulla, line 84: Probably don't want staging as default right? Here and on line 93

tashmulla (changes requested):
Just a couple comments about staging artefacts & config being left in, otherwise all good 🙂

tashmulla (approved):

PR #150, opened 2022-04-20
fix(): fix for missing ads on Unity Android

thomas-james-1986:
This PR fixes an issue where Ads wouldn't load on Unity Android. This was caused by using the `fromOrdinal` method which relies on the index position in the `SAConfiguration` of `Production` or `Staging` but the `SAConfiguration` in the Android SDK has another case `Dev` which was causing a mismatch. It seems to be safer overall to use the `fromValue` method which matches on the enum `rawValue` rather than the index position.

### DIFF
--- a/superawesome-base/src/test/java/tv/superawesome/lib/sasession/defines/SAConfigurationTest.java
+++ b/superawesome-base/src/test/java/tv/superawesome/lib/sasession/defines/SAConfigurationTest.java
@@ -9,6 +9,10 @@ import java.util.List;
 
 public class SAConfigurationTest  {
 
+    private static final int DEV = -1;
+    private static final int PRODUCTION = 0;
+    private static final int STAGING = 1;
+
     @Test
     public void test_ensure_parcel_and_un_parcel_works_dev(){
         // arrange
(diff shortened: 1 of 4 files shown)


superawesome-base/src/test/java/tv/superawesome/lib/sasession/defines/SAConfigurationTest.java
```diff
@@ -51,4 +51,40 @@ public void test_ensure_parcel_and_un_parcel_works_Production(){
         assertEquals(expected,actual);
     }
 
+    @Test
+    public void test_ensure_fromValue_returns_expected_config_Dev(){
+        // arrange
+        SAConfiguration expected = SAConfiguration.DEV;
+
+        // act
+        SAConfiguration actual = SAConfiguration.fromValue(-1);
```
tashmulla, line 60: Nice tests! Shall we move these integers out to variables so we know what they represent?

thomas-james-1986: Good shout! 👍

thomas-james-1986: Updated!

tashmulla (approved):
Nice one!

PR #157, opened 2022-05-24
fix: loading an ad with multiple parameters

gnhn:


thomas-james-1986 (approved):

tashmulla (approved):
Nice 👏 now that the endpoint is ready server-side, were you able to test the changes locally?

PR #158, opened 2022-05-24
Fix/aag 2076 asset file requests not published correctly

thomas-james-1986:
This is a first pass at fixing the broken documentation page. 
Some internal navigation (sidebar) still doesn't appear to be fixed locally but it may just be a local issue

Screenshot 2022-05-24 at 21 25 41

Screenshot 2022-05-24 at 21 44 29

### DIFF
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -9,14 +9,14 @@ ruby RUBY_VERSION
 #
 # This will help ensure the proper Jekyll version is running.
 # Happy Jekylling!
-# gem "jekyll", "3.2.1"
+gem "jekyll", "4.2.2"
 
 # This is the default theme for new Jekyll sites. You may change this to anything you like.
 # gem "minima"
 
 # If you want to use GitHub Pages, remove the "gem "jekyll"" above and
 # uncomment the line below. To upgrade, run `bundle update github-pages`.
-gem "github-pages", group: :jekyll_plugins
+# gem "github-pages", group: :jekyll_plugins
 
 # If you have any plugins, put them here!
 # group :jekyll_plugins do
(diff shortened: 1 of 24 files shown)


docs/Gemfile
```diff
@@ -9,14 +9,14 @@ ruby RUBY_VERSION
...
 # Happy Jekylling!
-# gem "jekyll", "3.2.1"
+gem "jekyll", "4.2.2"
 
 # This is the default theme for new Jekyll sites. You may change this to anything you like.
 # gem "minima"
 
 # If you want to use GitHub Pages, remove the "gem "jekyll"" above and
 # uncomment the line below. To upgrade, run `bundle update github-pages`.
-gem "github-pages", group: :jekyll_plugins
```
tashmulla, line 19: This was just for local development right?

docs/_includes/scripts.html
```diff
@@ -1,3 +1,3 @@
 <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.3/umd/popper.min.js" integrity="sha384-ZMP7rVo3mIykV+2+9J3UJ46jBk0WLaUAdn689aCwoqbBJiSnjAK/l8WvCWPIPm49" crossorigin="anonymous"></script>
 <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/js/bootstrap.min.js" integrity="sha384-ChfqqxuZUCnJSK3+MXmPNIyE6ZbWh2IMqE241rYiqJxyMiZ6OW/JmZQ5stwEULTy" crossorigin="anonymous"></script>
-<script src="{{ site.baseurl }}/assets/js/main.js"></script>
+<script src="{{ '/assets/js/' | relative_url }}main.js"></script>
```
tashmulla, line 3: Ooh interesting, how come this one is `/assets/js/` as opposed to just `/assets/` similar to the rest of the changes? (Same question for the change in `docs/pages/search.html` 🙂

tashmulla:
Nice one! To get the rest of the pages working, do you think it's a case of swapping out the use of `base_url` elsewhere? Might be good we do that anyway to keep consistency, but I can assume that could be an issue with the navigation?

docs/_includes/scripts.html
thomas-james-1986: Ah this was just due to the path where the file lives, most of them were in `assets` but some were in subfolders

thomas-james-1986:
@tashmulla yes I think so, I'll remove `base_url` everywhere to wrap this up

docs/_includes/scripts.html
tashmulla: Although e.g. in `head.html`, the files lived in either the `/favicons` or `/css` subfolders as well right?

thomas-james-1986: hmm yeah true, I think I need to match the folder structure everywhere?

tashmulla: I don't think it matters too much, as long as `/assets` is there as the first folder to look into - just probably makes sense to have it consistent everywhere, so would just go with whichever way makes more sense for you 🙂

gnhn:
If `baseUrl` is broken then how does it work for our other repositories?

PR #160, opened 2022-05-27
Feature/aag 2015 additional options

gnhn:
Android SDK: Allow optional config options to be passed through when initialising SDK to track against events


tashmulla:
Looks like something needs fixing up with the `android: exported` change according to the CI workflow. There's just a couple of these that are shown:

```
Error:
	android:exported needs to be explicitly specified for element <activity#androidx.test.core.app.InstrumentationActivityInvoker$BootstrapActivity>. Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined.
```

app/src/main/java/tv/superawesome/demoapp/MainActivity.kt
```diff
@@ -0,0 +1,169 @@
+package tv.superawesome.demoapp
+
+import android.app.Activity
+import android.os.Bundle
+import android.util.Log
+import android.view.View
+import android.widget.AdapterView
+import android.widget.AdapterView.OnItemClickListener
+import kotlinx.android.synthetic.main.activity_main.*
+import tv.superawesome.demoapp.adapter.*
```
thomas-james-1986, line 10: Not a biggie, but wondering if we need to import all here?

app/src/main/java/tv/superawesome/demoapp/MainActivity.kt
```diff
@@ -0,0 +1,169 @@
...
+        PlacementItem("Interstitial via KSF", 84799, type = Type.INTERSTITIAL),
+        HeaderItem("Videos"),
+        PlacementItem("Direct Video", 82090, type = Type.VIDEO),
+        PlacementItem(
+            "Vast Video",
+            84777,
+            lineItemId = 178822,
+            creativeId = 503585,
+            type = Type.VIDEO
+        ),
```
thomas-james-1986, line 32: 👍

app/src/main/java/tv/superawesome/demoapp/MainActivity.kt
```diff
@@ -0,0 +1,169 @@
...
+                                SAInterstitialAd.load(item.placementId, this@MainActivity)
+                            }
+                        }
+                        Type.VIDEO -> {
+                            if (SAVideoAd.hasAdAvailable(item.placementId)) {
+                                Log.i(TAG, "PLAYING VIDEO")
+                                SAVideoAd.play(item.placementId, this@MainActivity)
+                            } else {
+                                Log.i(TAG, "LOADING VIDEO")
+                                if (item.isFull()) {
```
thomas-james-1986, line 121: 👍

app/src/main/java/tv/superawesome/demoapp/MyApplication.kt
```diff
@@ -0,0 +1,11 @@
+package tv.superawesome.demoapp
+
+import androidx.multidex.MultiDexApplication
+import tv.superawesome.sdk.publisher.AwesomeAds
+
+class MyApplication : MultiDexApplication() {
+    override fun onCreate() {
+        super.onCreate()
+        AwesomeAds.init(this, true, mapOf("openRtbPartnerId" to "xyz123"))
```
tashmulla, line 9: Should we remove the 3rd parameter (I assume it was for testing)? Just if users download our SDK and try the demo app, they might wonder for that ID is 🙂

gnhn: Sure

tashmulla (approved):

PR #166, opened 2022-06-22
Feat/aag 2148 remove hardcoded version number

thomas-james-1986:
This PR updates the `SAVersion` version number to read from a `version.properties` file (if a version override isn't set). The next step is to modify `version.properties` in CI so the build number is updated automatically.

Added a screenshot to show the version number loaded from the properties file: 

Screenshot_20220623_162507


superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SAVersion.java
```diff
@@ -1,18 +1,46 @@
...
+     *
+     * @param context The context
      *
      * @return string representing the current version
      */
-    private static String getVersion() {
-        return version;
+    private static String getVersion(Context context) {
+        try {
+            return versionOverride == null ? loadVersion(context) : versionOverride;
```
gnhn, line 23: `getVersion` is called in every request/event. So It would be important that we cache the value, and don't read the file multiple times

superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SAVersion.java
```diff
@@ -1,18 +1,46 @@
...
+     * Method to load the current version from version.properties.
+     *
+     * @param context The context
+     *
+     * @return the stored string representing the current version
+     */
+
+    private static String loadVersion(Context context) throws IOException {
+        Properties properties = new Properties();
+        InputStream inputStream = context.getAssets().open("version.properties");
```
gnhn, line 41: We could potentially use `Resources.getSystem().getAssets()` here which uses the application context so we don't need to pass context here. 

`getVersion()` could be in use for `Admob` plugin and some Unity bridge methods so ideally we should keep it as it is -if possible-

gnhn: Though, maybe application context can't reach that file since it is inside this module. But would be nice if we can try

thomas-james-1986: Interesting, yeah it would be good to avoid passing the context if we can. I'll try it

thomas-james-1986: hmm, it isn't loading with `Resources.getSystem().getAssets()` unfortunately. I think we can pass the context from Unity, can we pass it from admob?

superawesome-base/src/main/resources/version.properties
```diff
@@ -0,0 +1 @@
+version.name=8.3.5
```
gnhn, line 1: Couldn't mark the above file but it seems it is empty
`superawesome-base/src/main/java/tv/superawesome/sdk/publisher/version.properties` 
And this one is filled
`superawesome-base/src/main/resources/version.properties`

thomas-james-1986: Great spot, this should be deleted. I Tried a few different locations

tashmulla: @thomas-james-1986 does this need to be 8.3.6 now? Just because 8.3.5 was released earlier by Gunhan to fix the Unity Android crashes bug

tashmulla (approved):

gnhn (approved):

PR #171, opened 2022-06-29
Placement updates

thomas-james-1986:
This PR adds the placements used within the automation testing to the example app


app/src/main/java/tv/superawesome/demoapp/MainActivity.kt
```diff
@@ -21,7 +21,9 @@ class MainActivity : Activity() {
         PlacementItem("Mobile Interstitial Flat Colour Portrait", 87892, type = Type.INTERSTITIAL),
         PlacementItem("Mobile Interstitial Portrait", 82089, type = Type.INTERSTITIAL),
         PlacementItem("Interstitial via KSF", 84799, type = Type.INTERSTITIAL),
+        PlacementItem("Interstitial Flat Colour via KSF", 87970, type = Type.INTERSTITIAL),
```
gnhn, line 24: Could you check that all UI tests are running correctly? As I remember I was set ui test to click a row and the index of the item could change with this.

thomas-james-1986: Sure I can check, I can potentially move it over to an XPath selector if there's an issue

thomas-james-1986: great spot 👍

PR #173, opened 2022-07-08
add: unit tests for the refactored code

gnhn:
- Created unit tests for refactored codebase
- Added `MainActivity2` for the refactored codebase


app/src/main/java/tv/superawesome/demoapp/MainActivity2.kt
```diff
@@ -0,0 +1,159 @@
...
+                        }
+                    }
+                    Type.VIDEO -> {
+                        if (SAVideoAd.hasAdAvailable(item.placementId)) {
+                            Log.i(TAG, "PLAYING VIDEO")
+                            SAVideoAd.play(item.placementId, this@MainActivity2)
+                        } else {
+                            SAVideoAd.load(item.placementId, this@MainActivity2)
+                        }
+//                        if (SAVideoAd.hasAdAvailable(item.placementId)) {
```
tashmulla, line 106: Was this commented-out block meant to be left in here?

app/src/main/java/tv/superawesome/demoapp/MainActivity2.kt
```diff
@@ -0,0 +1,159 @@
...
+            Log.i(TAG, "Config 2 selected")
+            bannerView.disableParentalGate()
+            bannerView.disableBumperPage()
+
+            SAInterstitialAd.disableParentalGate()
+            SAInterstitialAd.disableBumperPage()
+
+            SAVideoAd.disableParentalGate()
+            SAVideoAd.disableBumperPage()
+            // SAVideoAd.enableCloseButtonNoDelay()
```
thomas-james-1986, line 57: Should this be commented out here?

gnhn: Yes, unfortunately the new code doesn't have that functionality. But, we can keep it commented out until it is added maybe?

app/src/main/java/tv/superawesome/demoapp/MainActivity2.kt
gnhn: Yes, we don't have loading by lineItem, creativeId function at this moment. Missing feature on the refactor. Will be added and then we can uncomment.

thomas-james-1986, line 106: Do we still need these commented out lines?

gnhn: We can remove them and add them later when it is implemented, or just leave them for now

thomas-james-1986 (approved):

PR #174, opened 2022-07-13
Feat/aag 2193 mock server poc

thomas-james-1986:
This PR adds POC setup (not production) for Vapor and WireMock mock servers in order to show the pros / cons of each approach (having a standalone server in the case of Vapor and a more tightly integrated mock server in the case of WireMock).

I've added an interstitial UI test for each approach to show how they each might look in practice and a comment describing the main differences / pros and cons in each.

If you want to try the tests out, for Vapor you will need to install Vapor: https://docs.vapor.codes/install/macos/ and navigate to the root of the mock server folder (bundled here for PR review purposes but it would be separate in production) and run `vapor run`. You'll also need to uncomment the `UITESTING_URL` commented with "Vapor" in `SASession`.

For WireMock you will just need to uncomment the `UITESTING_URL` commented with "WIREMOCK". 

Again, this is POC code but I hope it helps us decide on an approach that we like. It's worth mentioning that there is also already `MockWebServer` in the project but this is only for Android I think and it only works with the particular networking library we're using, still we should consider this too since it's already available.

Screenshot of Vapor running:

Screenshot 2022-07-13 at 23 24 11


app/src/androidTest/java/tv/superawesome/demoapp/WireMockPOCTest.kt
```diff
@@ -0,0 +1,74 @@
...
+
+		/* With WireMock the server configuration and stubbing is done within the test files,
+			this has the advantage of being more flexible at the per test level. This is just a POC that
+			gets a response loading from the server but there are many things we could do to
+			build a scalable and clean setup. Mocks should be loaded from local JSON files for one. WireMock is also
+			compatible with iOS so we'd benefit from a familiar API. Again we need to point the app to localhost.
+		*/
+
+		stubFor(
+			get(urlPathMatching("/ad/87892"))
```
tashmulla, line 43: This setup looks pretty nice 👌  great suggestion in having local JSON files to define the mocks we want to use for each response type.

sa-sdk-mock-server/Dockerfile
```diff
@@ -0,0 +1,66 @@
+# ================================
```
tashmulla, line 1: This is useful to see!

PR #176, opened 2022-07-15
fix(): approval step in the circle ci

gnhn:


.circleci/config.yml
```diff
@@ -158,13 +158,17 @@ workflows:
   version: 2.1
   android_sdk_workflow:
     jobs:
+      - hold:
```
tashmulla, line 161: Could this be renamed to e.g. `approve_release` so it's clear on what approving the step will do? And then it would be good to have it only run after `test_superawesome_base` has passed (i.e. passing `test_superawesome_base` into the approval step's `requires` block), otherwise the steps may look a bit weird. Similar to how ads-server's set up is [here](https://github.com/SuperAwesomeLTD/sa-ads-server/blob/develop/.circleci/config.yml#L271)?

tashmulla: If we can move the approval job _below_ the `test_superawesome_base` job as well, that would make things clearer too I think, so then the order is more apparent - tests to pass, then approval stage, then deploy if approval given

.circleci/config.yml
```diff
@@ -158,13 +158,17 @@ workflows:
...
+          type: approval
```
tashmulla, line 162: Once we are happy with the approval step logistics, should we also add a filter block so it only runs on the `develop` branch? Otherwise we'll have forever-running pipelines in circleCI for feature branches as it'll be waiting for an approval for nothing technically 😄  So e.g.
```
filters:
  branches:
    only: develop
```

.circleci/config.yml
gnhn: Thanks for the suggestions, have a look again.

tashmulla:
Nice! Almost there - I think we just need one more update for the `semantic_release` `requires` section - right now it only runs on develop, which is good, but it only waits for `test_superawesome_base` to complete. Can we add in the `approve_release` to the `semantic_release` `requires` section please underneath `test_superawesome_base`? So then we ensure `approve_release` is required for:
- `prep_deploy`
- `semantic_release`

And then pushing to S3 is only done if `prep_deploy` runs, so I think that makes everything safe 🙂

gnhn:
> Nice! Almost there - I think we just need one more update for the `semantic_release` `requires` section - right now it only runs on develop, which is good, but it only waits for `test_superawesome_base` to complete. Can we add in the `approve_release` to the `semantic_release` `requires` section please underneath `test_superawesome_base`? So then we ensure `approve_release` is required for:
> 
> * `prep_deploy`
> * `semantic_release`
> 
> And then pushing to S3 is only done if `prep_deploy` runs, so I think that makes everything safe 🙂

I think this seems a bit circler referenced as `semantic_release` already have `test_superawesome_base` and we added `test_superawesome_base` to `approve_release` as well. Seems some part of this is redundant and we shouldn't need two different approvals to release as it currently releases without any approval.

tashmulla:
> > Nice! Almost there - I think we just need one more update for the `semantic_release` `requires` section - right now it only runs on develop, which is good, but it only waits for `test_superawesome_base` to complete. Can we add in the `approve_release` to the `semantic_release` `requires` section please underneath `test_superawesome_base`? So then we ensure `approve_release` is required for:
> > 
> > * `prep_deploy`
> > * `semantic_release`
> > 
> > And then pushing to S3 is only done if `prep_deploy` runs, so I think that makes everything safe 🙂
> 
> I think this seems a bit circler referenced as `semantic_release` already have `test_superawesome_base` and we added `test_superawesome_base` to `approve_release` as well. Seems some part of this is redundant and we shouldn't need two different approvals to release as it currently releases without any approval.

True - how about just getting rid of `test_superawesome_base` from the `semantic_release` requires? So then it'll go:
- Tests run no matter what
- Only tests to pass required for release approval
- Only release approval required for prep deploy & semantic release

gnhn:
> > > Nice! Almost there - I think we just need one more update for the `semantic_release` `requires` section - right now it only runs on develop, which is good, but it only waits for `test_superawesome_base` to complete. Can we add in the `approve_release` to the `semantic_release` `requires` section please underneath `test_superawesome_base`? So then we ensure `approve_release` is required for:
> > > 
> > > * `prep_deploy`
> > > * `semantic_release`
> > > 
> > > And then pushing to S3 is only done if `prep_deploy` runs, so I think that makes everything safe 🙂
> > 
> > 
> > I think this seems a bit circler referenced as `semantic_release` already have `test_superawesome_base` and we added `test_superawesome_base` to `approve_release` as well. Seems some part of this is redundant and we shouldn't need two different approvals to release as it currently releases without any approval.
> 
> True - how about just getting rid of `test_superawesome_base` from the `semantic_release` requires? So then it'll go:
> 
> * Tests run no matter what
> * Only tests to pass required for release approval
> * Only release approval required for prep deploy & semantic release

Let's add approval either to `semantic_release` or `prep_deploy` not both. As It adds no value to double approve for the same thing 😕

gnhn:
> > > Nice! Almost there - I think we just need one more update for the `semantic_release` `requires` section - right now it only runs on develop, which is good, but it only waits for `test_superawesome_base` to complete. Can we add in the `approve_release` to the `semantic_release` `requires` section please underneath `test_superawesome_base`? So then we ensure `approve_release` is required for:
> > > 
> > > * `prep_deploy`
> > > * `semantic_release`
> > > 
> > > And then pushing to S3 is only done if `prep_deploy` runs, so I think that makes everything safe 🙂
> > 
> > 
> > I think this seems a bit circler referenced as `semantic_release` already have `test_superawesome_base` and we added `test_superawesome_base` to `approve_release` as well. Seems some part of this is redundant and we shouldn't need two different approvals to release as it currently releases without any approval.
> 
> True - how about just getting rid of `test_superawesome_base` from the `semantic_release` requires? So then it'll go:
> 
> * Tests run no matter what
> * Only tests to pass required for release approval
> * Only release approval required for prep deploy & semantic release

I think now it is correct.
The workflow has correct requirements:
1 - test_superawesome_base executed
2 - approve_release -> waits for test_superawesome_base
3 - prep_deploy -> waits for approve_release
4 - semantic_release -> waits for prep_deploy
5 - sdk_push_superawesome
sdk_push_superawesome_base
sdk_push_superawesome_admob
-> waits for prep_deploy

The tests are executed always first, and approval is required before `prep_deploy` which runs the tests

tashmulla:
@gnhn looks good! So prep_deploy will only run after release approval, and then semantic release & S3 pushes will only run after prep_deploy. Looks much cleaner

tashmulla (approved):
👏

thomas-james-1986 (approved):
🚀

PR #178, opened 2022-08-03
add(): adding test cases for ad loading and also ad failing scenarios

gunhansancar:
- Adds test cases for ad loading for all ad types
- Adds test cases for ad failing for all ad types
- Adds a label to display ad loading result
image

- Added a new screen `Settings` to enable/disable `SDK` settings from the UI and from the UI tests.
image


app/src/androidTest/java/tv/superawesome/demoapp/util/WireMockHelper.kt
```diff
@@ -14,6 +14,17 @@ object WireMockHelper {
         )
     }
 
+    fun stubFailure(placement: String) {
+        stubFor(
+            get(urlPathMatching("/ad/$placement"))
+                .willReturn(
+                    aResponse()
+                        .withStatus(400)
```
tashmulla, line 22: This is still a 200 right? When we try and load a placement which doesn't exist, we get a 200 response and body of `{"response":"Placement not found."}`, e.g. here: https://ads.superawesome.tv/v2/ad/999999

tashmulla: (Specifically in the scenario we send an `adEmpty` callback anyway right?)

gunhansancar: I believe that is a different scenario for `adEmpty` cases where the `isValid` returns false in case of `Placement not found`.
So I will add a new test case for those scenarios 👍

app/src/androidTest/java/tv/superawesome/demoapp/util/ColorMatcher.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.hamcrest.Description
+import org.hamcrest.Matcher
+
+
+object ColorMatcher {
+    fun matchesColor(expectedColor: Color): Matcher<View> =
+        object : BoundedMatcher<View, View>(View::class.java) {
+            override fun matchesSafely(item: View): Boolean {
+                val capturedColor = ScreenshotUtil.captureColorInCenter(item)
+                return capturedColor == expectedColor
```
thomas-james-1986, line 15: Is there any flakiness here or is it able to pick the exact colour every time?

gunhansancar: Picks the exact color every time.

app/src/androidTest/java/tv/superawesome/demoapp/util/TestColors.kt
```diff
@@ -0,0 +1,11 @@
+package tv.superawesome.demoapp.util
+
+import android.graphics.Color
+
+object TestColors {
+    val yellow = Color.valueOf(0.96862745f, 0.8862745f, 0.41960785f)
```
thomas-james-1986, line 6: Do these different yellows help us identify the different ad types e.g. KSF / VAST and VPAID?

app/src/androidTest/java/tv/superawesome/demoapp/util/WireMockHelper.kt
```diff
@@ -14,6 +14,17 @@ object WireMockHelper {
...
+                        .withBody("")
+                )
+        )
+    }
+
     fun stubCommonPaths() {
```
thomas-james-1986, line 28: Nice, WireMock FTW

app/src/androidTest/java/tv/superawesome/demoapp/BannerUITest.kt
```diff
@@ -0,0 +1,45 @@
...
+import tv.superawesome.demoapp.util.ColorMatcher.matchesColor
+import tv.superawesome.demoapp.util.TestColors
+import tv.superawesome.demoapp.util.ViewTester
+import tv.superawesome.demoapp.util.waitUntil
+
+@RunWith(AndroidJUnit4::class)
+@SmallTest
+class BannerUITest {
+    @get:Rule
+    var wireMockRule = WireMockRule(8080)
```
thomas-james-1986, line 21: I wonder if there is a way to save repetition of some of the WireMock boilerplate, maybe a base test class could be useful?

gunhansancar: We can potentially move the `Rule` to a base class but I don't know if we will end up more test classes.

thomas-james-1986: 👍 maybe just something for us to keep in mind as the suite grows

app/src/androidTest/java/tv/superawesome/demoapp/util/TestColors.kt
gunhansancar: Yes, of course :) The test cases checks exact colour in each test. Any slight variation would cause test failure

thomas-james-1986 (approved):
👍 Pre-approving - just needs the scenarios Tash mentioned I think

tashmulla (approved):

PR #180, opened 2022-08-09
Build/circle ci semantic release dry run

tashmulla:
This PR adds a semantic release "dry run" step to our circleCI build & release process. It is set up to only run on `develop` so that we can ensure we are checking the semantic release version of the most up-to-date code before we approve to release.

The dry run will analyse the commits since the last released version and output the changelog and expected version to be released, or whether there just won't be a release. The dry run flag ensures we do not actually release the version outputted. This will allow us to double-check a version which will be released, or make a commit to ensure a particular release is made if the dry run says nothing will be released yet.

For context:
- semantic-release docs: https://www.businesswire.com/news/home/20220809005530/en/AppLovin-Submits-a-Compelling-Non-Binding-Proposal-to-Combine-with-Unity
- KWS PR adding similar implementation: https://github.com/SuperAwesomeLTD/freekws-nestjs-http-interceptors/pull/128/files
- Example semantic_release dry-run (KWS): https://app.circleci.com/pipelines/github/SuperAwesomeLTD/freekws-nestjs-http-interceptors/1050/workflows/5805ca19-2cb1-4c5a-a31c-cd0bef6e678e/jobs/3624

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -176,7 +176,7 @@ workflows:
           filters:
             tags:
               ignore: /^v.*/
-      -semantic_release_dry_run:
+      - semantic_release_dry_run:
           context: MOBILE_ANDROID_SDKS
           requires:
             - test_superawesome_base
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,7 @@ jobs:
   semantic_release_dry_run:
     <<: *android_config
     environment:
-      SA_SEMANTIC_RELEASE_RUN_BRANCH: ${CIRCLE_BRANCH}
+      SA_SEMANTIC_RELEASE_RUN_BRANCH: develop
     steps:
       - checkout
       - *generate_github_token
@@ -57,7 +57,7 @@ jobs:
           name: Check semantic release version
           command: |
             git clone https://x-access-token:${GITHUB_TOKEN}@github.com/SuperAwesomeLTD/sa-continuous-integration.git "${SA_PATH_TO_CONTINUOUS_INTEGRATION_REPO}"
-            ${SA_PATH_TO_CONTINUOUS_INTEGRATION_REPO}/semantic-release/semantic-release.sh --dry-run --branches ${CIRCLE_BRANCH}
+            ${SA_PATH_TO_CONTINUOUS_INTEGRATION_REPO}/semantic-release/semantic-release.sh --dry-run
 
   semantic_release:
     <<: *android_config
@@ -180,6 +180,9 @@ workflows:
           context: MOBILE_ANDROID_SDKS
           requires:
             - test_superawesome_base
+          filters:
+            branches:
+              only: develop          
       - approve_dry_run_output:
           type: approval
           requires:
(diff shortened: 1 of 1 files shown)


tashmulla:
> Looks good but I wonder if this displays the the next release version on somewhere and how?

@gunhansancar so it would be displayed but within the semantic-release run step output e.g. [here](https://app.circleci.com/pipelines/github/SuperAwesomeLTD/freekws-nestjs-http-interceptors/987/workflows/097f44a4-93eb-4b03-b4c3-110aabf2a5ac/jobs/3295/parallel-runs/0/steps/0-103) is an example of where it outputs the version at the bottom of the script. (Adding screenshot in case you aren't able to view the pipeline)

Screenshot 2022-08-09 at 16 47 37


Note that this would only run on `develop`, so post PR merge - so a PR would therefore be merged, dry run would output what release it would be (if any) after analysing the commits since the last tag, and the engineer would check this output to validate what will be released.

Would this suffice in terms of displaying the next release version?

thomas-james-1986:
Nice! Similar question to Gunhan, would we be able to grab the outputted version as a value to potentially bump it in the app?

tashmulla:
> Nice! Similar question to Gunhan, would we be able to grab the outputted version as a value to potentially bump it in the app?

@thomas-james-1986 ah gotcha, forgot about the version file we have! That's a good point. I believe we can use the nextRelease object which gets generated with a version property. I'll have a play around with it 🙂

thomas-james-1986:
> > Nice! Similar question to Gunhan, would we be able to grab the outputted version as a value to potentially bump it in the app?
> 
> @thomas-james-1986 ah gotcha, forgot about the version file we have! That's a good point. I believe we can use the nextRelease object which gets generated with a version property. I'll have a play around with it 🙂

Awesome! 🙌  I spotted this which might be what we'd need if we wanted to auto increment the version number in the sdk: https://github.com/Kerizer/fastlane-plugin-properties 🤔

tashmulla:
> > > Nice! Similar question to Gunhan, would we be able to grab the outputted version as a value to potentially bump it in the app?
> > 
> > 
> > @thomas-james-1986 ah gotcha, forgot about the version file we have! That's a good point. I believe we can use the nextRelease object which gets generated with a version property. I'll have a play around with it 🙂
> 
> Awesome! 🙌 I spotted this which might be what we'd need if we wanted to auto increment the version number in the sdk as a next step: https://github.com/Kerizer/fastlane-plugin-properties

Ah nice spot! That'll be something pretty good for us to explore 👌 

I had a look into the `nextRelease` object, and to e.g. use it to print out the version number to the pipeline, we'll need to make a change to the SA continuous-integration script which everyone uses [here](https://github.com/SuperAwesomeLTD/sa-continuous-integration/blob/master/semantic-release/semantic-release.sh). I think we'd benefit from getting the Platform team's help with figuring it out and doing it properly, and hopefully tying it in with fastlane, so are we happy to leave this PR as is (to just get dry-run working) and then have some follow-up work we can do with Platform to use the version number further? cc @thomas-james-1986 @gunhansancar

thomas-james-1986 (approved):

gunhansancar (approved):

PR #184, opened 2022-08-16
Chore/aag 2255 safe ad logo appearing multiple times

thomas-james-1986:
This PR adds a new serialiser for the `Ad` object and the logic to set `showPadlock` to `false` if `isKSF` is true as per: https://github.com/SuperAwesomeLTD/sa-mobile-sdk-android/commit/9e55ef4cbe36598491df03b5484f6b25843a005d along with unit tests. 

It seems like `showPadlock` isn't actually used in the refactored code at the moment 🤔 we seem to be using `Config.default` instead?  I can connect it up in this PR but I wanted to raise the PR at this point to get feedback on the changes so far.

### DIFF
--- a/superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/serializers/AdSerializer.kt
+++ b/superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/serializers/AdSerializer.kt
@@ -10,129 +10,129 @@ import tv.superawesome.sdk.publisher.common.models.Creative
 import tv.superawesome.sdk.publisher.common.models.CreativeDetail
 import tv.superawesome.sdk.publisher.common.models.CreativeFormatType
 
-class AdSerializer: KSerializer<Ad> {
+class AdSerializer : KSerializer<Ad> {
 
-	override val descriptor: SerialDescriptor = buildClassSerialDescriptor("Ad") {
-		element<Int?>("advertiserId")
-		element<Int>("publisherId")
-		element<Float>("moat")
-		element<Boolean>("is_fill")
-		element<Boolean>("is_fallback")
-		element<Int>("campaign_type")
-		element<Int>("campaign_id")
-		element<Boolean>("is_house")
-		element<Boolean>("safe_ad_approved")
-		element<Boolean>("show_padlock")
-		element<Int>("line_item_id")
-		element<Boolean>("test")
-		element<Int>("app")
-		element<String>("device")
-		element<Creative>("creative")
-	}
+    override val descriptor: SerialDescriptor = buildClassSerialDescriptor("Ad") {
+        element<Int?>("advertiserId")
+        element<Int>("publisherId")
+        element<Float>("moat")
+        element<Boolean>("is_fill")
+        element<Boolean>("is_fallback")
+        element<Int>("campaign_type")
+        element<Int>("campaign_id")
+        element<Boolean>("is_house")
+        element<Boolean>("safe_ad_approved")
+        element<Boolean>("show_padlock")
+        element<Int>("line_item_id")
+        element<Boolean>("test")
+        element<Int>("app")
+        element<String>("device")
+        element<Creative>("creative")
+    }
 
-	override fun serialize(encoder: Encoder, value: Ad)
-	{
-		encoder.encodeStructure(descriptor) {
+    override fun serialize(encoder: Encoder, value: Ad) {
+        encoder.encodeStructure(descriptor) {
 
-			// Optionals
-			value.advertiserId?.let { encodeIntElement(descriptor, 0, value.advertiserId) }
+            // Optionals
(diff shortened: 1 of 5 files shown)


superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/serializers/AdSerializer.kt
```diff
@@ -0,0 +1,138 @@
...
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.descriptors.buildClassSerialDescriptor
+import kotlinx.serialization.descriptors.element
+import kotlinx.serialization.encoding.*
+import tv.superawesome.sdk.publisher.common.models.Ad
+import tv.superawesome.sdk.publisher.common.models.Creative
+import tv.superawesome.sdk.publisher.common.models.CreativeDetail
+import tv.superawesome.sdk.publisher.common.models.CreativeFormatType
+
+class AdSerializer : KSerializer<Ad> {
```
gunhansancar, line 13: I'm not sure if we need a custom serialiser just for one boolean statement which can be added as an extension method to the `Ad` object. This adds extra complexity and it is very error prone. 

E.g.
`fun showPadlock(): Boolean = showPadlock && creative.isKSF`

I feel it is much cleaner.

thomas-james-1986: I know what you mean but then we'd have `showPadlock` and `showPadlock()` with different values which would be confusing. This way we could also change `isCPICampaign()` to a property

gunhansancar: When we add new values it will be always confusing and for the new developers this will be always error prone.

Better to handle custom logic in `AdResponse` also for the feature parity with iOS.

I understand the effort here but the sequentially ordering the elements one by one is not the way to go. 

We can open this dialog to wider developer groups though.

gunhansancar:
> This PR adds a new serialiser for the `Ad` object and the logic to set `showPadlock` to `false` if `isKSF` is true as per: [9e55ef4](https://github.com/SuperAwesomeLTD/sa-mobile-sdk-android/commit/9e55ef4cbe36598491df03b5484f6b25843a005d) along with unit tests.
> 
> It seems like `showPadlock` isn't actually used in the refactored code at the moment 🤔 we seem to be using `Config.default` instead? I can connect it up in this PR but I wanted to raise the PR at this point to get feedback on the changes so far.

That one seems to be missing.
I believe the best place to handle that is inside the `AdController`
It has been done the following way in iOS
`var showPadlock: Bool { adResponse?.advert.showPadlock ?? false && adResponse?.advert.ksfRequest == nil }`

We can add a new method to `AdResponse` as a function actually since the parameters come from there.
e.g.
```
data class AdResponse(
    val placementId: Int,
    val ad: Ad,
    var html: String? = null,
    var vast: VastAd? = null,
    var baseUrl: String? = null,
    var filePath: String? = null,
    var referral: String? = null,
) {
    fun isVideo(): Boolean = ad.creative.format == CreativeFormatType.Video

    fun showPadlock(): Boolean = ad.shouldShowPadlock()
```
And the `shouldShowPadlock` can be done this way inside the `Ad` object
`fun shouldShowPadlock(): Boolean = showPadlock && !creative.isKSF`

Please also note the we should get the opposite of `isKSF` here

superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/serializers/AdSerializer.kt
gunhansancar: Custom serialisers are to handle different types or formats. And, it is better to handle custom logic after the serialisation/deserialisation is done. We can use a mapper to convert DTO object to another object and handle the logic there or better yet `AdResponse` is there to handle that kind of implementations.

It could be even better we do that inside the `AdProcessor` as you can see we already do this kind of operations there.

tashmulla: Just weighing in my thoughts - I'd suggest we try and keep implementation as simple as possible for now and think about adding a custom serialiser later on if we agree? So I do think it does make more sense to just have e.g. `fun showPadlock(): Boolean = showPadlock && !creative.isKSF`, and then later on we can refactor this to either use a serialiser or just rename variables to avoid having `showPadlock()` and `showPadlock`?

We could keep this serialiser & test on a branch anyway and look to re-add it in later if we want to, so nothing is wasted?

thomas-james-1986: Ok understood - I'll close this PR and move the logic to `AdResponse` or `AdProcessor`.

PR #185, opened 2022-08-17
chore(): add timestamp to the requests

gunhansancar:
Adding the timestamp to the requests in the new codebase. 

Note: Java 8+ `Clock` is not preferred as a source for`TimeProvider` as it is only available API 26 and above. To be able to use it below API 26 we need to enable `coreLibraryDesugaring` which increases the size of our library.

Another alternative is `Calendar.getInstance(). getTimeInMillis()` This could be used but I wanted to avoid creating the `Calendar` object each time. However, if we think this can be better, then I can update the code.

### DIFF
--- a/superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/AdQueryMaker.kt
+++ b/superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/AdQueryMaker.kt
@@ -21,6 +21,7 @@ class AdQueryMaker(
     private val encoder: EncoderType,
     private val json: Json,
     private val locale: Locale,
+    private val timeProvider: TimeProviderType,
 ) : AdQueryMakerType {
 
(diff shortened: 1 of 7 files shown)


thomas-james-1986:
Hmmm I suppose we either take a hit to the size of the library or to performance using `Calendar.getInstance().getTimeInMillis()`. My personal preference would be to use `Calendar.getInstance().getTimeInMillis()` as I think the performance hit would be very negligible but it's quite close.

gunhansancar:
> Hmmm I suppose we either take a hit to the size of the library or to performance using `Calendar.getInstance().getTimeInMillis()`. My personal preference would be to use `Calendar.getInstance().getTimeInMillis()` as I think the performance hit would be very negligible but it's quite close.

Updated by using the `Calendar`

thomas-james-1986 (approved):
👍

PR #186, opened 2022-08-17
chore(): configuration for hiding the safe ad logo for KSF ads

thomas-james-1986:
This PR hides the safe ad logo for KSF Ads and fixes a crashing issue with the chucker library for API 31 and above in the common `build.gradle`.

I had to make some changes to `VideoActivity` since the `currentAdResponse` is not retrieved from the `AdStore` until `play`. The changes are similar to how the `WebView` is added in `BannerView` without the removal since the `VideoActivity` replaces the `IVideoPlayerControllerView` itself and it's destroyed on close. I've also matched the safe ad logo padding to the `BannerView` padding but we may want to tweak the padding / positioning for both for styling reasons.


gunhansancar (approved):
👍

PR #189, opened 2022-08-22
Fix/aag 2263 get version from properties

thomas-james-1986:
This PR adds support for loading the `version.properties` file to `SdkInfo`. Whilst adding this I noticed that the `overridePlatform` String isn't used, do we need to make sure this class can output the same SDK version string as `SAVersion.getSDKVersion()` ?

### DIFF
--- a/superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/SdkInfo.kt
+++ b/superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/SdkInfo.kt
@@ -17,7 +17,12 @@ class SdkInfo(
     private val encoder: EncoderType,
     locale: Locale
 ) : SdkInfoType {
-    private object VersionInfo {
+    private object Keys {
+        const val Unknown = "unknown"
+        const val Platform = "android"
+        const val VersionFile = "version.properties"
+        const val VersionName = "version.name"
+
         val storedVersion = loadVersion()
 
         /**
@@ -34,8 +39,8 @@ class SdkInfo(
     }
 
     override val version: String
-        get() = overrideVersion ?: "${Platform}_${VersionInfo.storedVersion}"
-    override val bundle: String = context.packageName ?: Unknown
+        get() = overrideVersion ?: "${Keys.Platform}_${Keys.storedVersion}"
+    override val bundle: String = context.packageName ?: Keys.Unknown
     override val name: String by lazy { findAppName() }
     override val lang: String = locale.toString()
 
@@ -43,7 +48,7 @@ class SdkInfo(
         val label = context.packageManager.getApplicationLabel(context.applicationInfo).toString()
         encoder.encodeUri(label)
     } catch (exception: PackageManager.NameNotFoundException) {
-        Unknown
+        Keys.Unknown
     }
 
     companion object {
@@ -51,8 +56,3 @@ class SdkInfo(
         var overridePlatform: String? = null
     }
 }
-
-const val Unknown = "unknown"
-const val Platform = "android"
-const val VersionFile = "version.properties"
-const val VersionName = "version.name"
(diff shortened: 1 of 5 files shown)


superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/SdkInfo.kt
```diff
@@ -14,16 +15,33 @@ interface SdkInfoType {
...
         const val unknown = "unknown"
         const val platform = "android"
+        val storedVersion = loadVersion()
+
+        /**
+         * Method to load the current version from version.properties.
+         *
+         * @return the stored string representing the current version
+         */
+
```
gunhansancar, line 30: Formatting: extra line here can be removed

superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/SdkInfo.kt
```diff
@@ -14,16 +15,33 @@ interface SdkInfoType {
...
+        fun loadVersion(): String {
```
gunhansancar, line 31: Could be this way
```
    private const val VersionFile = "version.properties"
    private const val VersionName = "version.name"

    fun loadVersion(): String {
        val inputStream = javaClass.classLoader?.getResourceAsStream(VersionFile) ?: return ""
        val properties = Properties()
        properties.load(inputStream)
        return properties.getProperty(VersionName)
    }
```

Constant could go to the bottom of the file

superawesome-common/src/test/java/tv/superawesome/sdk/publisher/common/components/SdkInfoTest.kt
```diff
@@ -26,7 +26,7 @@ class SdkInfoTest : BaseTest() {
         }
         val encoderType = Encoder()
 
-        val sdkInfo = SdkInfo(context, encoderType, locale, "1.2.3")
+        val sdkInfo = SdkInfo(context, encoderType, locale)
```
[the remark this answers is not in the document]
thomas-james-1986: The assert on line 35 tests it although I didn't change the line, it's loading the 1.2.3 from the version.properties file resource

superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/SdkInfo.kt
thomas-james-1986: Yes that's cleaner. Do you mean move the constants out of the `object`?

gunhansancar: Better to not use companion object for the constants which creates extra getters/setters when compiled. And the access will be the same anyway. So we can prefer to move them to the bottom of the file out of object. What do you think?

superawesome-common/src/test/java/tv/superawesome/sdk/publisher/common/components/SdkInfoTest.kt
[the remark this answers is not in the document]
gunhansancar: Perfect

superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/SdkInfo.kt
thomas-james-1986: Sure sounds good. FYI I kept the `?:` after `getProperty` as it can return `null` if the key doesn't exist

gunhansancar (approved):

superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/components/SdkInfo.kt
thomas-james-1986: Actually I think these constants are ok IMHO, they're on an `object` as opposed to a companion object

PR #191, opened 2022-08-23
Chore/aag 2262 show close button immediately

thomas-james-1986:
This PR adds the immediate close button feature for the refactor


superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/state/CloseButtonState.kt
```diff
@@ -0,0 +1,13 @@
+package tv.superawesome.sdk.publisher.common.state
+
+enum class CloseButtonState(val value: Int) {
+    Hidden(0), VisibleWithDelay(1), VisibleImmediately(2);
+
+    fun isVisible(): Boolean =
```
gunhansancar, line 6: Would be nice if we add test cases for `isVisible` and `fromInt` functions

thomas-james-1986: Yes I agree, I'll add some tests around the state

gunhansancar (approved):

PR #195, opened 2022-08-31
Chore/aag 2388 investigate ad ended and ad closed duplicate events

thomas-james-1986:
This PR adds a fix for multiple `adEnded` and `adClosed` events being sent. It was already hinted at in the comments in the code - resetting on complete fixes it. There was also a call to `removeTimer()` within reset so there doesn't seem to be a need to call it twice. Since it's safe to call I've moved it out of the `try` in `reset()`.

Before:

Screenshot 2022-08-31 at 12 32 44

After:

Screenshot 2022-08-31 at 12 42 03

### DIFF
--- a/superawesome-base/src/main/java/tv/superawesome/sdk/publisher/videoPlayer/VideoPlayerController.kt
+++ b/superawesome-base/src/main/java/tv/superawesome/sdk/publisher/videoPlayer/VideoPlayerController.kt
@@ -104,14 +104,12 @@ class VideoPlayerController :
     }
 
     override fun onCompletion(mediaPlayer: MediaPlayer) {
-        //removeTimer()
         reset()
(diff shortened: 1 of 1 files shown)


gunhansancar:
I found a side effect when I run with these changes.
Steps:
1- I click the Flat colour video ad
2- Then when the remaining time is 4 seconds and minimise the app
3- After 4 seconds ad is closed while app is minimised
Expected:
Ad should continue where it is left(Current behaviour on the `develop` branch)

thomas-james-1986:
> I found a side effect when I run with these changes. Steps: 1- I click the Flat colour video ad 2- Then when the remaining time is 4 seconds and minimise the app 3- After 4 seconds ad is closed while app is minimised Expected: Ad should continue where it is left(Current behaviour on the `develop` branch)

Hmm ok nice catch, I'll investigate

PR #199, opened 2022-09-05
fix(): triggering semantic release on dry run approval

thomas-james-1986:
This PR triggers `semantic_release ` on `approve_dry_run_and_release` which subsequently triggers `prep_deploy`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -193,7 +193,7 @@ workflows:
       - prep_deploy:
           context: MOBILE_ANDROID_APPS
           requires:
-            - approve_dry_run_and_release
+            - semantic_release
           filters:
             branches:
               ignore: /.*/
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -202,7 +202,7 @@ workflows:
       - semantic_release:
           context: MOBILE_ANDROID_SDKS
           requires:
-            - prep_deploy
+            - approve_dry_run_and_release
```
tashmulla, line 205: I think you'll need to swap round `prep_deploy` and `semantic_release` order since `prep_deploy` needs `semantic_release` to run first?

tashmulla: So after approval is triggered, semantic_release is run (and a tag is released) and then prep_deploy & subsequent jobs are run on this tag workflow

thomas-james-1986: Ah yes, that should be fixed now

tashmulla (approved):
🚀

PR #204, opened 2022-09-06
Feat/aag 2246 auto version bump test 1

thomas-james-1986:
This PR is an initial test to see if we can log out the next release version from the semantic-release script


tashmulla (approved):
👍

PR #206, opened 2022-09-06
Feat/aag 2246 auto version bump test 3

thomas-james-1986:
This PR attempts outputting nextRelease data via a workspace

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,6 +68,8 @@ jobs:
   output_next_release_version:
     <<: *android_config
     steps:
+      - attach_workspace:
+          at: .
       - run:
           name: Output next release
           command: echo nextRelease
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ jobs:
     <<: *android_config
     steps:
       - run:
-          name: Output next release version
+          name: Output next release
           command: echo nextRelease
 
   semantic_release:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,8 +67,6 @@ jobs:
 
   output_next_release_version:
     <<: *android_config
-    environment:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -51,15 +51,26 @@ jobs:
...
+          root: "."
+          paths:
+          - "nextRelease"
+
+  output_next_release_version:
+    <<: *android_config
+    steps:
+      - run:
+          name: Output next release
+          command: echo nextRelease
```
tashmulla, line 73: Does this need to be e.g. `${nextRelease}` as it's a variable? Not 100% sure but feels like it may just print out the string itself?

.circleci/config.yml
```diff
@@ -51,15 +51,26 @@ jobs:
...
-            echo ${SA_PATH_TO_CONTINUOUS_INTEGRATION_REPO}/semantic-release/semantic-release.sh --dry-run ${nextRelease.version}
+      - persist_to_workspace:
+          root: "."
+          paths:
+          - "nextRelease"
+
+  output_next_release_version:
+    <<: *android_config
+    steps:
+      - run:
```
tashmulla, line 71: Does this need an `attach_workspace` too in order to access the workspace? Again, not 100% sure on this!

.circleci/config.yml
thomas-james-1986: Not sure, I was referencing these docs: https://circleci.com/docs/workspaces which uses `cat /tmp/workspace/echo-output` but perhaps, there's a bit of trial and error here!

.circleci/config.yml
thomas-james-1986: That would make sense, it crossed my mind too, I'll add it

.circleci/config.yml
tashmulla: Oooh interesting, I wonder if `cat` may work a little different to `echo`, but yeah, could start from here and we know to add the variable syntax if we see `nextRelease` as a string is just echoed out!

tashmulla (approved):

PR #213, opened 2022-09-07
chore(): Add tests for bumper page on Banner, Interstitial, and Video

gunhansancar:
- Add tests for bumper page on Banner, Interstitial, and Video
- Added new setting on the main screen to enable/disable bumper page


app/src/androidTest/java/tv/superawesome/demoapp/VideoAdUITest.kt
```diff
@@ -172,12 +193,61 @@ class VideoAdUITest {
...
+            .waitForView(withContentDescription("Ad content"))
+            .perform(waitUntil(isDisplayed()))
+            .perform(click())
+
+        // Then bumper page is shown
+        BumperInteraction.waitUntilBumper()
+
+        // And view URL is redirected to browser
+        Thread.sleep(4500)
+        intended(IntentMatchers.hasAction(Intent.ACTION_VIEW))
```
tashmulla, line 231: Is it possible to check for `/video/click` URL here? Is that what should be expected?

tashmulla: (I know this will technically be covered in the Events checking ticket, but just in case it's easy to do since you already checked for `/click` for banner and interstitial 😄)

gunhansancar: I tried to add that check but actually it is not called.

Further debugging shows that we don't actually trigger `video/click` event for videos at all (which would be normally called through `events.triggerClickEvent()` on the current codebase) that `click` event is only called via `Banner` and `Interstitial` since it uses `Banner` internally.

I don't know if this is the expected behaviour since we have VAST events like `vast_click_tracking` and `vast_click_through` which are triggered by code for videos.

tashmulla: Oh interesting! So do we trigger the VAST click events for direct video clicks at least? 😂  Or do we actually trigger zero click events for non-VAST videos right now? So intriguing as I do remember seeing the `/video/click` logs coming through earlier in the year, maybe we were imagining things 😅

tashmulla: Here's a report for our usual direct video placement (82090) and we have generated click events at least - hopefully it means we are doing this through vast click tracking at least? https://dashboard.superawesome.tv/en/reports/result/a86d1af7-7610-440e-863d-75229a8abb5a

gunhansancar: This may sound unintuitive but all of the video ads have vast tag except `vpaid` (even we call them direct video). Thus, we trigger vast click events for any of the video ads.

tashmulla: Ah okay good to know!

Regarding the logs, I may have been imagining it 😂  I think we're good for now since we're tracking `click` events in some fashion, that's the main outcome we want. Let's keep this in mind though to sort out as it does confuse things if someone isn't aware that all video ads do actually have a vast tag

tashmulla: In that case, should we add a test for the vast click tracking in this PR then since we do technically trigger it? Happy for it also to be covered next sprint too 🙂

gunhansancar: Also, we have `/video/click` trigger on iOS so those numbers may come from there

tashmulla: We do get _some_ click events from Android, maybe it's to do with the amount we even tested clicks on Android: https://dashboard.superawesome.tv/en/reports/result/a5188b2c-176a-4b1c-ba8d-512e0e3e6b70

tashmulla (approved):

thomas-james-1986 (approved):

PR #215, opened 2022-09-16
Feat/aag 2413 common events UI tests

thomas-james-1986:
This PR adds automated tests around common events for the different ad types, specifically:

**Banners:**

Impressions
Viewable impressions
Click

**Interstitials:**

Impressions
Viewable Impressions
Click

**Direct Videos**

Impressions (via VAST)
Viewable Impression
Click (via VAST)
Dwell Time

**VAST Videos**

Impressions
Viewable Impressions
Click
Dwell Time

Also, using `sleep` is a bit flaky especially for viewable impressions as we're waiting for the asset to load and the wait time for a viewable impression to trigger. I think we can improve this by having the mock server serve the mp4 assets as then they should load straight away, I thought I'd raise the PR as is though. Perhaps it can be a separate ticket? Or I can do it as part of this one.


thomas-james-1986:
> Regarding this:
> 
> > I can't see a click event or a normal impression for direct videos, although it seems to be sending a VAST impression?
> > Shall I create separate bug tickets for these?
> 
> * Sending VAST impression: @gunhansancar did we speak about this recently that because all videos have a `vast` property anyway so we therefore send a VAST impression for direct video?
> * No click on direct video: @thomas-james-1986 is this just not implemented at all in the code, even if e.g. an actual button exists in the direct video in order to click through to a webpage?

I've just updated the direct video impressions test to see if it sends a VAST impression, I can confirm that it does since it has a VAST block as you say.

For the click, it looks like it runs through this code:

`private void handleUrl (Context context, String destination) {

        Long currentTime = System.currentTimeMillis()/1000;
        long diff = Math.abs(currentTime - currentClickThreshold);

        if (diff < SADefaults.defaultClickThreshold()) {
            Log.d("AwesomeAds-2", "Current diff is " + diff);
            return;
        }

        currentClickThreshold = currentTime;

        Log.d("AwesomeAds-2", "Going to " + destination);

        // send vast click tracking events
        events.triggerVASTClickTrackingEvent();

        // send only in case of CPI where we'll use the direct click url
        if (ad.campaignType == SACampaignType.CPI) {
            events.triggerVASTClickThroughEvent();
        }

        // if it's a CPI campaign
        destination += ad.campaignType == SACampaignType.CPI ? ("&referrer=" + ad.creative.referral.writeToReferralQuery()) : "";

        // start browser
        try {
            context.startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse(destination)));
        } catch (Exception e) {
            // do nothing
        }
    }`

I can't see anything there for a direct video click, could it be related to the VAST block again?

app/src/androidTest/java/tv/superawesome/demoapp/VideoAdUITest.kt
```diff
@@ -20,7 +20,9 @@ import tv.superawesome.demoapp.interaction.BumperInteraction
 import tv.superawesome.demoapp.interaction.CommonInteraction
```
tashmulla, line 20: How do we record events etc. for VPAID ads? Do we trigger anything from our mobile SDK e.g. by ensuring impression URLs are hit? Or is everything handled on the web SDK side?

thomas-james-1986: Good question, I couldn't see anything being triggered for VPAID. I'll have another look 👀

tashmulla: Thanks! I know we do record events for VPAID in Metabase (Druid) when playing the ad from Android, so something _must_ be triggering somewhere... Speaking of, Metabase access is also a good thing for you to ask IT for as well when you get a chance 😄

thomas-james-1986: Do you know how VPAID events are handled Gunhan? It seems like they might be handled on the web SDK side?

gunhansancar: For `vpaid` events we do nothing. Everything is handled behind the scenes. We have callbacks though for some of the events e.g. `adFinished` `adClosed` etc.

thomas-james-1986: Nice one, thanks Gunhan, that seemed to be the case 🙌

tashmulla: Thanks both!

tashmulla (approved):
LGTM!

PR #216, opened 2022-09-16
chore(): adds new uitests for parental gate events

gunhansancar:
chore(): adds new uitests for parental gate events

### DIFF
--- a/app/src/androidTest/java/tv/superawesome/demoapp/util/ParentalGateUtil.kt
+++ b/app/src/androidTest/java/tv/superawesome/demoapp/util/ParentalGateUtil.kt
@@ -1,0 +1,11 @@
+package tv.superawesome.demoapp.util
+
+object ParentalGateUtil {
+    fun solve(inputText: String): Int {
+        val split = inputText.split("+")
+        if (split.size != 2) return 0
+        val number1 = split[0].replace(("[^\\d.]").toRegex(), "").toInt()
+        val number2 = split[1].replace(("[^\\d.]").toRegex(), "").toInt()
+        return number1 + number2
+    }
+}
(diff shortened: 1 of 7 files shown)


app/src/androidTest/java/tv/superawesome/demoapp/util/ParentalGateUtil.kt
```diff
@@ -0,0 +1,11 @@
+package tv.superawesome.demoapp.util
+
+object ParentalGateUtil {
+    fun solve(inputText: String): Int {
```
thomas-james-1986, line 4: 😂 👍

app/src/androidTest/java/tv/superawesome/demoapp/util/WireMockHelper.kt
```diff
@@ -102,4 +102,8 @@ object WireMockHelper {
         verify(anyRequestedFor(urlPathMatching(urlPath))
             .withQueryParam(queryParamKey, matching(queryParamValueRegex)))
     }
+
+    fun verifyQueryParamContains(urlPath: String, paramKey: String, paramValue: String) {
+        verify(anyRequestedFor(urlPathMatching(urlPath)).withQueryParam(paramKey, containing(paramValue)))
```
thomas-james-1986, line 107: Nice, this looks a little similar to the one I added above but I think this one is cleaner at the call site because we may not need to use regex? If the common event tests work with this method of inspecting the query params perhaps we can swap `verifyUrlPathCalledWithQueryParam` with it?

gunhansancar: One uses regex one uses contains so maybe we can keep both for now for the tests

thomas-james-1986: Ok fair enough 👍

thomas-james-1986 (approved):
👍

PR #221, opened 2022-09-28
Feat/aag 2249 test callbacks

thomas-james-1986:
This PR adds UI tests for the following callbacks (with a few questions - **X** means added). 
I swapped out `checkSubtitle` with `checkSubtitleContains` so we can use it to check the history of callbacks (the callback name gets appended to the label's text)

Note: I added an `adClosed` event on the close button in `SAManagedAdActivity` as it wasn't being sent on the action. Is that correct?

**Banner:**

adLoaded X
adEmpty X
adFailedToLoad X
adAlreadyLoaded: Can this be called for banners?
adShown X
adFailedToShow: It looks like this is called in the case of `Web_Error` but I'm not sure how to replicate this?
adClicked X
adEnded N/A
adClosed X

**Interstitial:**

adLoaded X
adEmpty X
adFailedToLoad X
adAlreadyLoaded - X
adShown X
adFailedToShow - Not sure if this is called?
adClicked X
adEnded X
adClosed X

**Video**

adLoaded X
adEmpty X
adFailedToLoad X
adAlreadyLoaded - X
adShown X (Not called for VPAID?)
adFailedToShow - Not sure if this is called?
adClicked X
adEnded X
adClosed X

### DIFF
--- a/app/src/androidTest/java/tv/superawesome/demoapp/BannerUITest.kt
+++ b/app/src/androidTest/java/tv/superawesome/demoapp/BannerUITest.kt
@@ -16,12 +16,12 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import tv.superawesome.demoapp.interaction.BumperInteraction
 import tv.superawesome.demoapp.interaction.CommonInteraction
-import tv.superawesome.demoapp.interaction.CommonInteraction.stubIntents
 import tv.superawesome.demoapp.interaction.ParentalGateInteraction
 import tv.superawesome.demoapp.interaction.SettingsInteraction
 import tv.superawesome.demoapp.util.ColorMatcher.matchesColor
 import tv.superawesome.demoapp.util.TestColors
 import tv.superawesome.demoapp.util.ViewTester
+import tv.superawesome.demoapp.util.WireMockHelper.stubIntents
 import tv.superawesome.demoapp.util.WireMockHelper.verifyUrlPathCalled
 import tv.superawesome.demoapp.util.WireMockHelper.verifyUrlPathCalledWithQueryParam
 import tv.superawesome.demoapp.util.isVisible
--- a/app/src/androidTest/java/tv/superawesome/demoapp/BannerUITest.kt
+++ b/app/src/androidTest/java/tv/superawesome/demoapp/BannerUITest.kt
@@ -150,6 +150,23 @@ class BannerUITest {
             .check(isVisible())
     }
 
+    @Test
+    fun test_adClosed_callback() {
+        val placement = "88001"
+        CommonInteraction.launchActivityWithSuccessStub(placement, "banner_success.json")
+
+        CommonInteraction.clickItemAt(2)
+
+        ViewTester()
+            .waitForView(withId(R.id.bannerView))
+            .perform(waitUntil(matchesColor(TestColors.yellow)))
+
+        CommonInteraction.clickItemAt(2)
+
+        // The banner is closed automatically when a new one is opened
+        CommonInteraction.checkSubtitleContains("$placement adClosed")
+    }
+
     // Events
     @Test
     fun test_banner_impression_events() {
--- a/app/src/androidTest/java/tv/superawesome/demoapp/BannerUITest.kt
+++ b/app/src/androidTest/java/tv/superawesome/demoapp/BannerUITest.kt
@@ -21,7 +21,7 @@ import tv.superawesome.demoapp.interaction.SettingsInteraction
 import tv.superawesome.demoapp.util.ColorMatcher.matchesColor
 import tv.superawesome.demoapp.util.TestColors
 import tv.superawesome.demoapp.util.ViewTester
-import tv.superawesome.demoapp.util.WireMockHelper.stubIntents
+import tv.superawesome.demoapp.util.IntentsHelper.stubIntents
 import tv.superawesome.demoapp.util.WireMockHelper.verifyUrlPathCalled
 import tv.superawesome.demoapp.util.WireMockHelper.verifyUrlPathCalledWithQueryParam
 import tv.superawesome.demoapp.util.isVisible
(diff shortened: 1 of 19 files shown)


superawesome-base/src/main/java/tv/superawesome/sdk/publisher/managed/SAManagedAdActivity.kt
```diff
@@ -44,7 +44,10 @@ class SAManagedAdActivity : Activity(), AdViewJavaScriptBridge.Listener {
         closeButton.setPadding(0, 0, 0, 0)
         closeButton.scaleType = ImageView.ScaleType.FIT_XY
         closeButton.layoutParams = buttonLayout
-        closeButton.setOnClickListener { close() }
+        closeButton.setOnClickListener {
+            listener?.onEvent(this.placementId, SAEvent.adClosed)
```
gunhansancar, line 48: Should we call `adClosed` inside the `close()` function as it can be called in multiple locations? Not sure though

app/src/main/java/tv/superawesome/demoapp/MainActivity.kt
```diff
@@ -222,7 +222,8 @@ class MainActivity : FragmentActivity() {
     }
 
     private fun updateMessage(placementId: Int, event: SAEvent) {
-        val message = "$placementId $event"
+        val originalMessage = subtitleTextView.text
+        val message = "$originalMessage $placementId $event"
```
gunhansancar, line 226: Wouldn't this recursively adds all the callbacks to the textview? I wonder if we should use a `listview` or something more elegant here? 🤷

app/src/androidTest/java/tv/superawesome/demoapp/util/WireMockHelper.kt
```diff
@@ -90,6 +95,15 @@ object WireMockHelper {
         )
     }
 
+    fun stubIntents() {
```
gunhansancar, line 98: We should move this to `IntentsHelper` or `IntentsUtil` as it is not related to `WireMock`

app/src/androidTest/java/tv/superawesome/demoapp/interaction/CommonInteraction.kt
```diff
@@ -58,6 +66,39 @@ object CommonInteraction {
...
             .check(matches(withText(text)))
     }
+
+    fun checkSubtitleContains(text: String, timeout: Long = 5000) {
+        onView(withId(R.id.subtitleTextView))
+            .perform(waitUntil(isDisplayed()))
+            .perform(waitForText(text, timeout))
+    }
+
+    fun waitForText(text: String, timeout: Long): ViewAction = object : ViewAction {
```
gunhansancar, line 79: Could we try to use `waitUntil` instead of this? I am not sure but this may work `waitUntil(withSubstring(""))`?

app/src/main/java/tv/superawesome/demoapp/MainActivity.kt
thomas-james-1986: It's not recursive, it stores the original message and appends to it. It's only for debugging, not sure it's worth spending the time building out a more elaborate non-user facing UI. The text box does expand with the text.

app/src/androidTest/java/tv/superawesome/demoapp/util/WireMockHelper.kt
thomas-james-1986: Ok good shout

app/src/androidTest/java/tv/superawesome/demoapp/interaction/CommonInteraction.kt
thomas-james-1986: I can try it, this does work reasonably well, I've re-checked all of the tests but if that works and we don't need to create a custom method that would be better. This at least avoids `Thread.sleep` and things like that.

superawesome-base/src/main/java/tv/superawesome/sdk/publisher/managed/SAManagedAdActivity.kt
thomas-james-1986: It crossed my mind too but it might yield duplicate events, I'll check

app/src/androidTest/java/tv/superawesome/demoapp/BannerUITest.kt
```diff
@@ -149,6 +150,23 @@ class BannerUITest {
             .check(isVisible())
     }
 
+    @Test
+    fun test_adClosed_callback() {
+        val placement = "88001"
+        CommonInteraction.launchActivityWithSuccessStub(placement, "banner_success.json")
+
+        CommonInteraction.clickItemAt(2)
```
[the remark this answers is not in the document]
thomas-james-1986: Good shout, that's much less brittle - so I should upgrade to `clickPlacementById` here we can fix across all the tests with your new convenience when it's available?

app/src/androidTest/java/tv/superawesome/demoapp/interaction/CommonInteraction.kt
thomas-james-1986: It worked 👌

superawesome-base/src/main/java/tv/superawesome/sdk/publisher/managed/SAManagedAdActivity.kt
thomas-james-1986: I've moved it to the close function inside the `(!isFinishing)` check as suggested in standup so it should be triggered once. FYI @tashmulla I think you mentioned `adEnded` and vpaid in standup, I changed our vpaid `adClosed` test locally to look for `adEnded` and possible duplicate events but I didn't see `adEnded`, I did see however the screen go white at the end of the video (when the timer hits zero)? Thought it could be useful info

app/src/main/java/tv/superawesome/demoapp/MainActivity.kt
gunhansancar: Hmm, then we can maybe figure out better UI later

gunhansancar (approved):
👍

PR #225, opened 2022-10-13
fix(): crash when the video is completed

gunhansancar:
fix(): crash when the video is completed but later on clicked causing…

### DIFF
--- a/superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SAVideoActivity.java
+++ b/superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SAVideoActivity.java
@@ -106,7 +106,9 @@ public class SAVideoActivity extends Activity implements IVideoPlayer.Listener, 
         chrome.setShouldShowSmallClickButton(config.shouldShowSmallClick);
         chrome.setClickListener(view -> {
             videoClick.handleAdClick(view);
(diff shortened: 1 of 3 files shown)


tashmulla:
Nice 👏  can I confirm we've checked the following:

- Crash does not appear anymore

- UI & unit tests pass

- Events & callbacks still being sent correctly (e.g. `adClosed` `adEnded` `adClicked` callbacks, server events) - this should all be covered in the UI tests right?

- Playback behaviour is as expected when auto-close enabled or disabled:
  - auto-close enabled:
    - video just auto-closes on finish
  - auto-close disabled:
    - video finishes and last frame visible with x button
    - if user minimises app and returns to app, video does not re-start and last frame stays visible with x button
    - user can close ad with x button

- click behaviour is as expected when auto-close enabled or disabled:
  - auto-close enabled:
    - user can click on ad and through to webpage many times during ad playback
  - auto-close disabled:
    - user can click on ad and through to webpage many times during ad playback
    - user can click on ad when its finished and through to webpage without ad crashing

- Playback behaviour is as expected for minimising ad in general:
  - video content pauses when app is minimised and plays from last place when going back to app
  - video sound pauses when app is minimised and plays from last place when going back to app

More of a sense-check than anything as we know it's a little bit of a change with logic 😄

thomas-james-1986 (approved):

PR #229, opened 2022-11-14
chore(): mute and unmute parameter for vpaid

gunhansancar:
chore(): mute and unmute parameter for vpaid


superawesome-base/src/main/java/tv/superawesome/lib/sasession/defines/SARTBPlaybackMethod.java
```diff
@@ -5,15 +5,17 @@
...
+    WITH_SOUND_OFF_SCREEN(2);
 
     private final int value;
 
     SARTBPlaybackMethod(int value) {
         this.value = value;
     }
 
-    public static SARTBPlaybackMethod fromValue (int skip) {
+    public static SARTBPlaybackMethod fromValue(int skip) {
```
thomas-james-1986, line 17: `skip` seems like a strange param name here?

thomas-james-1986 (approved):
Looks good, just a question about the skip param name

superawesome-base/src/main/java/tv/superawesome/lib/sasession/defines/SARTBPlaybackMethod.java
gunhansancar: Yes, agree. Updated to `playbackValue`

PR #232, opened 2022-11-23
chore(): update tests to use local files

gunhansancar:
chore(): update tests to use local files


app/src/androidTest/java/tv/superawesome/demoapp/BannerUITest.kt
```diff
@@ -31,7 +32,7 @@ import tv.superawesome.demoapp.util.waitUntil
 @SmallTest
 class BannerUITest {
     @get:Rule
-    var wireMockRule = WireMockRule(8080)
+    var wireMockRule = WireMockRule(wireMockConfig().port(8080), false)
```
thomas-james-1986, line 35: What does the false boolean control?

thomas-james-1986 (approved):
Looks good, just a query on the WireMock boolean

app/src/androidTest/java/tv/superawesome/demoapp/BannerUITest.kt
gunhansancar: It is to control `failOnUnmatchedRequest` checks.
That internal check for the request matching was giving false-positive failed tests when I run the whole suite of tests. 
Our verify checks for the requests still work.

thomas-james-1986: Nice 👍

PR #236, opened 2022-12-06
bugfix(): Fix the admob integration and use new admob interfaces

gunhansancar:


saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/SAAdMobAdapter.kt
```diff
@@ -0,0 +1,67 @@
...
+            VersionInfo(splits[0].toInt(), splits[1].toInt(), splits[2].toInt())
+        } else {
+            VersionInfo(0, 0, 0)
+        }
+    }
+
+    override fun getVersionInfo(): VersionInfo {
+        val splits = SAVersion.getSDKVersionNumber().split(".")
+        return if (splits.size >= 3) {
+            VersionInfo(splits[0].toInt(), splits[1].toInt(), splits[2].toInt() * 100)
```
thomas-james-1986, line 62: Just wondering what the expected output of this is? 🤔

saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/SAAdMobBannerCustomEvent.java
```diff
@@ -21,125 +21,128 @@
...
+                                    customEventBannerListener.onAdClosed();
+                                    break;
+
+                            }
+                        });
+
+        try {
+            bannerAd.addOnLayoutChangeListener(
+                    (v, left, top, right, bottom, oldLeft, oldTop, oldRight, oldBottom) -> {
+                        layouted = true;
```
thomas-james-1986, line 115: perhaps `layoutUpdated` ?

saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/SAAdMobAdapter.kt
gunhansancar: So here we provide our `admob` plugin version, usually the minor version is just played our nicely so that it reflect our default sdk version also it differs from it by multiplying the last digit. However, it is just common practice we can use the same version as AA sdk version

saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/SAAdMobBannerCustomEvent.java
gunhansancar: I directly refactored the old code with the existing variables. Yes, I can rename it one of the suggestions

thomas-james-1986 (approved):

PR #237, opened 2022-12-07
fix(): This PR prevents the app from going to sleep when playing video ads

thomas-james-1986:
This PR prevents apps using the SDK from entering sleep mode when video ads are playing. Device sleep is only prevented whilst the video activities are open, exiting the video screen means sleep mode can occur as normal.

### DIFF
--- a/superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SAVideoActivity.java
+++ b/superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SAVideoActivity.java
@@ -16,6 +16,7 @@ import android.os.Parcelable;
 import android.util.DisplayMetrics;
 import android.util.Log;
 import android.view.View;
+import android.view.WindowManager;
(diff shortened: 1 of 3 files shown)


superawesome-base/src/main/java/tv/superawesome/sdk/publisher/SAVideoActivity.java
```diff
@@ -66,6 +67,7 @@ public class SAVideoActivity extends Activity implements IVideoPlayer.Listener,
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
```
gunhansancar, line 70: Could you also add this to the `VideoActivity.kt`?

gunhansancar (approved):

PR #245, opened 2023-01-10
Feat/aag 2716 load placements from firebase

thomas-james-1986:
This PR moves the data source for the placements list to Firebase so that it's easier for people to update and maintain without needing a rebuild of the app

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,8 @@ references:
       name: Configure Google Services
       command: |
         echo "Google Services"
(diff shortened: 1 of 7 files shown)


tashmulla (commented):
(All good once Gunhan's comment addressed)

gunhansancar:
Please `squash and merge` and the title and message should be `chore()` as the changes doesn't occur in the main SDK as they're about the example app.

thomas-james-1986:
> (All good once Gunhan's comment addressed)

This is a slightly sensitive file (although not particularly) which is already gitignored so we've added instructions here for how to get it from Firebase when setting up: https://superawesomeltd.atlassian.net/wiki/spaces/AA/pages/4907696505/Android

gunhansancar (approved):

PR #247, opened 2023-01-12
Feature/aag 2707 admob update

gunhansancar:
Updating `admob` module to use the latest refactored codebase


saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/SAAdMobInterstitialCustomEvent.kt
```diff
@@ -0,0 +1,102 @@
...
+    "Kept for backward compatibility reasons",
+    ReplaceWith("SAAdMobAdapter")
+)
+class SAAdMobInterstitialCustomEvent : CustomEventInterstitial {
+    private var context: Context? = null
+    private var loadedPlacementId = 0
+    override fun requestInterstitialAd(
+        context: Context,
+        listener: CustomEventInterstitialListener,
+        s: String?,
```
thomas-james-1986, line 32: `s` is an unusual param name, it looks like it relates to a placementId on line 80?

gunhansancar: Nice catch. I didn't notice that parameter at all. Those lines are updated automatically by formatting.
I will rename it to `placementId`

saadmob/build.gradle
```diff
@@ -30,8 +30,8 @@ repositories {
 }
 
 dependencies {
-    api project(':superawesome-base')
-    implementation 'com.google.android.gms:play-services-ads:20.1.0'
-    implementation 'androidx.core:core-ktx:1.6.0-beta02'
+    api project(':superawesome-common')
+    implementation 'com.google.android.gms:play-services-ads:21.4.0'
+    implementation 'androidx.core:core-ktx:1.9.0'
```
me2k, line 35: 👍

saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/SAAdMobAdapter.kt
```diff
@@ -19,7 +19,7 @@ class SAAdMobAdapter : Adapter() {
         initializationCompleteCallback: InitializationCompleteCallback,
         mediationConfigurations: MutableList<MediationConfiguration>
     ) {
-        AwesomeAds.init(context, false)
+        AwesomeAdsSdk.init(context, Configuration())
```
me2k, line 22:
```suggestion
        AwesomeAdsSdk(context, Configuration())
```
Surely this is possible?

saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/SAAdMobBannerAd.kt
```diff
@@ -31,30 +29,29 @@ class SAAdMobBannerAd(
         loadedPlacementId =
             adConfiguration.serverParameters.getString(SAAdMobExtras.PARAMETER)?.toIntOrNull() ?: 0
 
-        bannerAd = SABannerAd(context)
-        bannerAd?.id = ID
+        bannerAd = BannerView(context)
```
me2k, line 32: What is the decision to drop the SA prefix? Is this part of modernising the code? As a consumer of this SDK I might create my own BannerView class thus causing ambiguity with this library. Adding our own prefix separates this concern and makes the integrating developer's job more pain less.

saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/SAAdMobBannerAd.kt
```diff
@@ -64,17 +61,17 @@ class SAAdMobBannerAd(
...
                 // do nothing
             }
-            SAEvent.adShown -> adCallback?.onAdOpened()
-            SAEvent.adFailedToShow -> adFailedToLoad()
-            SAEvent.adClicked -> adCallback?.reportAdClicked()
-            SAEvent.adClosed -> adCallback?.onAdClosed()
+            SAEvent.AdShown -> adCallback?.onAdOpened()
+            SAEvent.AdFailedToShow -> adFailedToLoad()
+            SAEvent.AdClicked -> adCallback?.reportAdClicked()
+            SAEvent.AdClosed -> adCallback?.onAdClosed()
```
me2k, line 74: 👍 Good to see some warnings silenced 😁

me2k (approved):
Great work, from my little knowledge on the code base so far all I see are good things in this PR. 👏

saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/SAAdMobAdapter.kt
gunhansancar: Not really, it is not a class constructor.

saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/SAAdMobBannerAd.kt
gunhansancar: Prefixing class names come from Objective-C era where the namespace system doesn't exist. It is not recommended in any of the modern languages as it is not necessary. It is highly unlikely that a screen displays multiple banner views from different libraries than they had to use full package name for one :) But I can't see in any world that would happen to our clients.

saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/SAAdMobBannerAd.kt
gunhansancar: The new `SAEvent` cases starts with uppercase as it is recommended by Kotlin conventions however to make things easier for the migration we can rename it later before doing the actual switch.

thomas-james-1986 (approved):
Nice one!

PR #248, opened 2023-01-17
NO-JIRA: Added PR Template

me2k:
Added PR Template


Banford (approved):
Makes sense to me. Not too heavy but adds a little structure. Thanks.

PR #250, opened 2023-01-18
chore(): refactor android unity plugin to use the new common module

gunhansancar:
## JIRA
[AAG-2708]

## DESCRIPTION
Refactoring unity plugin code to use new common module

[AAG-2708]: https://superawesomeltd.atlassian.net/browse/AAG-2708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/saunity/src/main/java/tv/superawesome/plugins/publisher/unity/SAUnityBannerAd.java
+++ b/saunity/src/main/java/tv/superawesome/plugins/publisher/unity/SAUnityBannerAd.java
@@ -13,11 +13,11 @@ import android.widget.FrameLayout;
 
 import java.util.HashMap;
 
-import tv.superawesome.lib.sasession.defines.SAConfiguration;
-import tv.superawesome.lib.sautils.SAUtils;
-import tv.superawesome.sdk.publisher.SABannerAd;
-import tv.superawesome.sdk.publisher.SAEvent;
-import tv.superawesome.sdk.publisher.SAInterface;
+import tv.superawesome.plugins.publisher.unity.util.SAScreenUtil;
+import tv.superawesome.sdk.publisher.common.components.NumberGenerator;
+import tv.superawesome.sdk.publisher.common.models.SAEvent;
+import tv.superawesome.sdk.publisher.common.ui.banner.BannerView;
+
 
 /**
  * Class that holds a number of static methods used to communicate with Unity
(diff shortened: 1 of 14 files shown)


saunity/src/main/java/tv/superawesome/plugins/publisher/unity/SAUnityBannerAd.java
```diff
@@ -13,40 +13,58 @@
...
     // hash map containing banner ads
-    private static final HashMap<String, SABannerAd> bannerAdHashMap = new HashMap<>();
+    private static final HashMap<String, BannerView> bannerAdHashMap = new HashMap<>();
 
     /**
      * Method that creates a new banner ad (from Unity)
      */
-    public static void SuperAwesomeUnitySABannerAdCreate (Context context, final String unityName) {
-
+    public static void SuperAwesomeUnitySABannerAdCreate(Context context, final String unityName) {
```
thomas-james-1986, line 33: Should we remove `SA` from this method too?

saunity/src/main/java/tv/superawesome/plugins/publisher/unity/SAUnityBannerAd.java
```diff
@@ -13,40 +13,58 @@
...
-                case adEmpty: SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.adEmpty.toString()); break;
-                case adFailedToLoad: SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.adFailedToLoad.toString()); break;
-                case adAlreadyLoaded: SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.adAlreadyLoaded.toString()); break;
-                case adShown: SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.adShown.toString()); break;
-                case adFailedToShow: SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.adFailedToShow.toString()); break;
-                case adClicked: SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.adClicked.toString()); break;
-                case adEnded: SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.adEnded.toString()); break;
-                case adClosed: SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.adClosed.toString()); break;
+                case AdLoaded:
+                    SAUnityCallback.sendAdCallback(unityName, placementId, SAEvent.AdLoaded.getValue());
```
thomas-james-1986, line 42: Should we keep `SA` here?

thomas-james-1986 (commented):
Looks good, just curious about the `SA` prefix in some places and whether we should still be doing it?

saunity/src/main/java/tv/superawesome/plugins/publisher/unity/SAUnityBannerAd.java
gunhansancar: We can but this names are called from Unity by method name in a string so we need to also update there. Kind of ok to change but then we need to test

saunity/src/main/java/tv/superawesome/plugins/publisher/unity/SAUnityVideoAd.java
```diff
@@ -24,34 +22,34 @@ public class SAUnityVideoAd {
      * Method that creates a new Video Ad (from Unity)
      */
     public static void SuperAwesomeUnitySAVideoAdCreate(Context context) {
```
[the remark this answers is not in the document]
gunhansancar: Maybe we should come up with better names as these are used from Unity scripts

saunity/src/main/java/tv/superawesome/plugins/publisher/unity/SAUnityBannerAd.java
thomas-james-1986: Yeah it would mean quite a few more changes. Perhaps it's a good time to get rid of it though?

me2k (approved):
This looks good to me, hence approval. Unsure on the SA naming convention here though. Is there an initiative to remove it?

thomas-james-1986 (approved):

PR #259, opened 2023-02-06
chore(): fix a bug where bumper page screen doesn't prevent clicks go through

gunhansancar:
## JIRA
[AAG-2806]

## DESCRIPTION
- Fix a bug where bumper page screen doesn't prevent the clicks go through. 
- Also converts bumper page to a dialog instead of an activity.

## 
Screenshot 2023-02-06 at 21 00 50
SCREENSHOTS

[AAG-2806]: https://superawesomeltd.atlassian.net/browse/AAG-2806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/ui/common/BumperPageDialog.kt
```diff
@@ -0,0 +1,120 @@
...
+
+    fun updateTimeLeft(timeLeft: Int) {
+        smallText?.text = context.getString(R.string.bumper_page_time_left, timeLeft)
+    }
+}
+
+private const val BIG_LOGO_ID = 0x1201012
+private const val SMALL_LOGO_ID = 0x212121
+private const val SMALL_TEXT_ID = 0x212151
+private const val BIG_TEXT_ID = 0x212751
```
me2k, line 120: Shouldn't these be in a companion object?

superawesome-common/src/main/res/values/strings.xml
```diff
@@ -8,4 +8,5 @@
     <string name="proceed">Continue</string>
     <string name="bumper_page_time_left">A new site will open in %1$d seconds. Remember to stay safe online and don’t share your username or password with anyone!</string>
     <string name="bumper_page_leaving"> Bye! You’re now leaving %1$s</string>
+    <string name="bumper_app_name_default">this app</string>
```
me2k, line 11: Change this to something more reasonable? Like `Example App`

me2k (approved):

superawesome-common/src/main/res/values/strings.xml
gunhansancar: We can't use `Example app` as this string is appended to the end of
`Bye! You're now leaving this app`

If you meant fetching the original app name from the package that's not requested before and they can override the name using 
`BumperPage.overrideName` function to fill any other name e.g their app name.

superawesome-common/src/main/java/tv/superawesome/sdk/publisher/common/ui/common/BumperPageDialog.kt
gunhansancar: It is actually better to keep constants as a file private location instead of Companion object because when kotlin companion object converted into bytecode, the file private becomes less intrusive.
Example
```
class BumperPageDialog {
    companion object {
        const val BIG_LOGO_ID = 0x1201012
    }
}
```
Becomes
```
public final class BumperPageDialog {
    public static final String BIG_LOGO_ID = 0x1201012;
    public static final BumperPageDialog.Companion Companion = new BumperPageDialog.Companion((DefaultConstructorMarker) null);

    public static final class Companion {
        private Companion() {}

        public Companion(DefaultConstructorMarker $constructor_marker) {
            this();
        }
    }
}
```

Whereas the file private const becomes
```
public final class BumperPageDialog {}

public final class BumperPageDialogKt {
   public static final String BIG_LOGO_ID = 0x1201012;
}

```
By the way, if we increase the min version number from `16` to `17` then we can just use
`View.generateViewId()` 
and remove those constants.

me2k: Wow, that is a really comprehensive answer, thank you for taking the time to share! That helps a lot.

PR #283, opened 2023-02-27
AAG-2879: Added timeouts to show and hide the close button on Managed Ads

me2k:
## JIRA
[AAG-2879]

## DESCRIPTION
This PR Adds timers to show the close button after a 2 second delay if configured to be visible with a delay and after a 12 second timeout if required when viewing a Managed (VPAID) video.

I have also fixed a couple of bugs in the Common module when testing. The missing version file has been added but names `newVersion.properties` to avoid conflicts.
Fixed an issue that prevented the VPAID ad closing when finished.
Fixed an issue that meant the `adShown` event was fired twice, once from the web bridge and once from the ad controller

## SCREENSHOTS

Close Button Visible Immediately

https://user-images.githubusercontent.com/618350/221842552-2d448ab4-2542-42b1-ba73-6f88fed0e89c.mov

Close Button Visible After Delay

https://user-images.githubusercontent.com/618350/221842522-e10dced4-5291-49e5-b785-a87383c51fe3.mov

Close Button Hidden

https://user-images.githubusercontent.com/618350/221842570-622cad2e-b276-4106-9502-d4ef7efe2d32.mov

Close Button 12 second fallback timeout

https://user-images.githubusercontent.com/618350/221842589-0f90cd84-d751-45ff-9cfb-6011e7111ab3.mov

[AAG-2879]: https://superawesomeltd.atlassian.net/browse/AAG-2879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


superawesome-base/src/main/java/tv/superawesome/sdk/publisher/managed/SAManagedAdActivity.kt
```diff
@@ -55,15 +66,59 @@ class SAManagedAdActivity : Activity(), AdViewJavaScriptBridge.Listener {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
+        // get values from the intent
+        val myIntent = intent
+        config = myIntent.getParcelableExtra(CONFIG_KEY)
```
gunhansancar, line 72: Let's simplify as `intent.getParcelableExtra(CONFIG_KEY)` no need to create a local variable

superawesome-common/src/main/resources/newVersion.properties
```diff
@@ -0,0 +1 @@
+version.name=8.5.3
```
thomas-james-1986, line 1: isn't this the version file the CI writes to? I think CI is looking for `version.properties` ?

me2k: This is not, when we remove the base module, this will be renamed `version.properties` and the correct "previous version" will be added here.

PR #299, opened 2023-03-17
fix(admob): Allow event listener to listen multiple ad loads

gunhansancar:
## JIRA
[AAG-2965]

## DESCRIPTION
fix(admob): Allow event listener for admob adapter to listen multiple ad loads at the same time

[AAG-2965]: https://superawesomeltd.atlassian.net/browse/AAG-2965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/EventListener.kt
+++ b/saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/EventListener.kt
@@ -1,0 +1,60 @@
+package tv.superawesome.plugins.publisher.admob
+
+import android.util.Log
+import tv.superawesome.sdk.publisher.SAEvent
+import tv.superawesome.sdk.publisher.SAInterface
+import java.lang.ref.WeakReference
+
+internal class EventListener : SAInterface {
+    private var observers: MutableList<WeakReference<SAInterface>> = mutableListOf()
+
+    fun subscribe(listener: SAInterface) {
+        observers.forEach {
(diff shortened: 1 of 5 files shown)


saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/EventListener.kt
```diff
@@ -0,0 +1,60 @@
...
+                    "EventListener subscribe -> Already subscribed"
+                )
+                return
+            }
+        }
+        observers.add(WeakReference(listener))
+        Log.d(
+            "SuperAwesome",
+            "EventListener subscribe -> Subscribed Total: ${observers.count()}"
+        )
```
me2k, line 26: Should these logs be kept behind a debug flag?

me2k (approved):
Awesome job! You've also kept backwards compatibility 😁

thomas-james-1986 (approved):

saadmob/src/main/java/tv/superawesome/plugins/publisher/admob/EventListener.kt
gunhansancar: Throughout the SDK the logs are done this way. The refactored code base handle the logs better. Once we start using it we can update this.

PR #314, opened 2023-03-31
fix(Bumper Page): Prevent clicks from registering

gunhansancar:
## JIRA
[AAG-3038]

## DESCRIPTION
fix(Bumper Page): Prevent clicks from registering when the user clicks outside the bumper page

## SCREENSHOTS
n/a

[AAG-3038]: https://superawesomeltd.atlassian.net/browse/AAG-3038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/build.gradle
```diff
@@ -67,6 +67,7 @@ dependencies {
     androidTestImplementation "androidx.test:core-ktx:1.4.0"
     androidTestImplementation "androidx.test.espresso:espresso-core:$espresso_core_version"
     androidTestImplementation "androidx.test.espresso:espresso-intents:$espresso_core_version"
+    androidTestImplementation 'androidx.test.uiautomator:uiautomator:2.2.0'
```
me2k, line 70: Interesting that you've added this, I added it in my UI Test CI branch but I didn't think it did anything.

app/src/androidTest/java/tv/superawesome/demoapp/BannerUITest.kt
```diff
@@ -20,13 +22,11 @@ import tv.superawesome.demoapp.interaction.CommonInteraction
 import tv.superawesome.demoapp.interaction.ParentalGateInteraction
 import tv.superawesome.demoapp.interaction.SettingsInteraction
 import tv.superawesome.demoapp.model.TestData
+import tv.superawesome.demoapp.util.*
```
me2k, line 25: Do you like wild card imports? If I were to turn on Detekt it has a default rule that rejects these...

me2k (approved):
Great job 👏  We really need to get rid of the base module. There are a lot of changes here 😬

app/build.gradle
gunhansancar: It is required for `UiDevice.getInstance(getInstrumentation()).click(0, 100)` which allows to directly interact with the device and click anywhere on the screen.

app/src/androidTest/java/tv/superawesome/demoapp/BannerUITest.kt
gunhansancar: I use auto import so this has to be done automatically I guess. I can update if you prefer

PR #318, opened 2023-04-24
fix(UITesting): Adding a test rule to retry the tests in order to reduce flakiness

thomas-james-1986:
## JIRA
[AAG-3027]

## DESCRIPTION

This PR adds a retrying test rule to reduce test failures. I'm also resetting the WireMock stubs before each test to clear off any per test stubs. 

This significantly improves the pass rate but I still see some failures which I think can be addressed by adding more stubs.


me2k (dismissed):
Good idea 😁 Let's see how this improves the test reliability.

thomas-james-1986:
> Good idea 😁 Let's see how this improves the test reliability.

33/35 on the last run

PR #320, opened 2023-04-24
fix(Video): Added Play Pause for Videos in the Base module

me2k:
## JIRA
[AASDK-343]

## DESCRIPTION
Added Play Pause for Videos in the Base module

## SCREENSHOTS

https://user-images.githubusercontent.com/618350/234048753-5afac924-8e64-4f11-8c8d-3a9b8c311d87.mov

[AASDK-343]: https://superawesomeltd.atlassian.net/browse/AASDK-343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gunhansancar (approved):

thomas-james-1986 (commented):
Looks good, just wanted to clarify a few behaviours around the parental gate after testing iOS a bit more:

a) on iOS tapping the continue button with an empty parental gate solution field throws the error message, I think on Android the parental gate gets dismissed here? Just wondering if they should be aligned.

b) I don't think the ad should unpause after successfully completing the parental gate (if it does at the moment) as the user gets sent to a link or to the bumper straight away and then the ad gets unpaused when the user returns to the app

c) on iOS we need to unpause on the `parentalGateFailure` event as that gets triggered instead of `parentalGateCancel` when getting the parental gate sum wrong, might be worth a double check of the flow on Android

me2k:
@thomas-james-1986 Here's a few answers and actions I have executed:

> a) on iOS tapping the continue button with an empty solution field throws the error message, I think on Android the parental gate gets dismissed here? Just wondering if they should be aligned.

This is certainly the case in the Android Base module, but not the common module. 
I've created [AASDK-448](https://superawesomeltd.atlassian.net/browse/AASDK-448) to deal with this.

> b) I don't think the ad should unpause after successfully completing the parental gate (if it does at the moment) as the user gets sent to a link or to the bumper straight away and then the ad gets unpaused when the user returns to the app

I don't think it should unpause either, I noticed this behaviour when working on the Android Common module and have corrected it where possible. I feel the ad should unpause when the app becomes active.

> c) on iOS we need to unpause on the `parentalGateFailure` event as that gets triggered instead of `parentalGateCancel` when getting the parental gate sum wrong, might be worth a double check of the flow on Android

This is correctly implemented in Android Base already and now in Android Common in this PR: [AASDK-445](https://github.com/SuperAwesomeLTD/sa-mobile-sdk-android/pull/321)

thomas-james-1986 (approved):

PR #327, opened 2023-04-28
chore(UITests): Extended existing UITests to check that a browser is opened with the correct URLs

me2k:
## JIRA
[AASDK-396]

## DESCRIPTION
Extended existing UITests to check that a browser is opened with the correct URLs when tapping the ad and that you can return to the app without crashing.

The tests run and pass by themselves, but when run as a suite some fail still.

The Video UI Tests now pass 100% every time, there were memory leaks, but these have been fixed. I would assume similar issues exist in the Interstitials and banners.

## SCREENSHOTS
n/a

[AASDK-396]: https://superawesomeltd.atlassian.net/browse/AASDK-396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


superawesome-base/src/main/java/tv/superawesome/sdk/publisher/managed/SAManagedAdActivity.kt
```diff
@@ -127,6 +127,9 @@ class SAManagedAdActivity : Activity(),
         adView.pauseVideo()
         cancelCloseButtonTimeoutRunnable()
         cancelCloseButtonShownRunnable()
+        listener = null
+        config = null
+        videoClick = null
```
me2k, line 132: This is to clean up memory leaks

superawesome-base/src/main/java/tv/superawesome/sdk/publisher/managed/AdViewJavaScriptBridge.kt
```diff
@@ -1,40 +1,49 @@
...
-    fun adPaused() = listener.adPaused()
+    fun adPaused() = tryListener { listener.adPaused() }
+
+    private fun tryListener(block: () -> Unit) {
+        try {
+            block()
+        } catch (e: Exception) {
+            Log.d("SuperAwesome", "JSBridge Error ${e.message}")
+        }
+    }
```
me2k, line 46: This cleans up the exception log

superawesome-base/src/main/java/tv/superawesome/sdk/publisher/managed/SAManagedAdActivity.kt
gunhansancar: `onStop` is called when the activity minimised as well like if they click Home button on the device and then return back the app. At that point the listener would be gone and they would never receive events afterwards. I don't think this is the right approach.

superawesome-base/src/main/java/tv/superawesome/sdk/publisher/managed/AdViewJavaScriptBridge.kt
gunhansancar: We should also reflect these changes inside the `common` since we will soon start to use it. In `common` let's use `Logger` instead of direct `Log`

superawesome-base/src/main/java/tv/superawesome/sdk/publisher/managed/SAManagedAdActivity.kt
```diff
@@ -125,9 +125,17 @@ class SAManagedAdActivity : Activity(),
     override fun onStop() {
         super.onStop()
         adView.pauseVideo()
+    }
+
+    override fun finish() {
+        super.finish()
+        listener = null
```
gunhansancar, line 132: `finish()` is not part of activity lifecycle as it needs to be called manually to destroy the activity in which inherently calls `onDestroy()` itself.

gunhansancar (approved):

PR #330, opened 2023-05-05
chore(AdMob): Update AdMob plugin to use common module

gunhansancar:
## JIRA
[AASDK-466]

## DESCRIPTION
chore(AdMob): Update AdMob plugin to use common module

## SCREENSHOTS
n/a

[AASDK-466]: https://superawesomeltd.atlassian.net/browse/AASDK-466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/tv/superawesome/demoapp/admob/AdMobActivity.kt
```diff
@@ -59,7 +60,7 @@ class AdMobActivity : Activity() {
     private fun requestInterstitialAd() {
         val bundle = SAAdMobExtras.extras()
             .setTestMode(false)
-            .setOrientation(SAOrientation.PORTRAIT)
+            .setOrientation(Orientation.Portrait)
```
me2k, line 63: Orientation is too generic a name, this might cause conflicts with publishers apps. `SAOrientation` is safer or make Orientation internal and redefine it for the test app.

me2k (approved):
Great job 👍

app/src/main/java/tv/superawesome/demoapp/admob/AdMobActivity.kt
gunhansancar: Since this is a compile time issue I don't think there would be any problems. They can't just use any other `Orientation` here. We shouldn't worry about this kind naming with the modern languages like `Kotlin`, `Java`, and `Swift` :)

PR #332, opened 2023-05-09
chore(): Added content descriptions to the settings UI for the test app

thomas-james-1986:
## JIRA
[[AASDK-476](https://superawesomeltd.atlassian.net/browse/AASDK-476)]

## DESCRIPTION

This PR adds content descriptions to the settings view for UI testing. iOS already has them in place. I'll create new binaries for iOS / Android for Sean when we've confirmed we've got everything we want to get into `develop` in. 

## SCREENSHOTS

[AASDK-476]: https://superawesomeltd.atlassian.net/browse/AASDK-476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/tv/superawesome/demoapp/main/MainActivity.kt
+++ b/app/src/main/java/tv/superawesome/demoapp/main/MainActivity.kt
@@ -49,6 +49,7 @@ class MainActivity : FragmentActivity() {
                 updateSettings()
             }
         }
+        settingsButton.contentDescription = "AdList.Buttons.Settings"
     }
 
     private fun initUI() {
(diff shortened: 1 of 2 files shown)


me2k (approved):
Awesome Job 👍

gunhansancar (approved):
Looks good 👍 

* We should not set random content descriptions for the actual ad views as these descriptions are read out load by the accessibility settings when enabled. So the close button should say `Close` etc. But of course this PR only affects the example app which is fine to use it that way.

